PR #8, opened 2019-10-17
feat: Add Experimental DirectPath support

igorbernstein2:
Previously DirectPath support was enabled only in a test profile. Now the support is being pulled into the actually client. This functionality is currently in private alpha testing and will only be available to some customers. A lot of the changes in this PR are temporary, and will be removed once DirectPath goes public using the default endpoint.

The endpoint hasn't rolled out yet, but I was able to confirm that this works with the old test endpoint. The new endpoint should be live tomorrow


kolea2:
lgtm after code format

PR #9, opened 2019-10-17
WIP: enable grpc stats when enabling Bigtable stats

igorbernstein2:
To get a full picture of metrics for the bigtable client, the customer would need to look at both operation metrics and grpc rpc metrics. This was previously disabled because grpc metrics had a bug that prevented the method and status tags from being populated. This was fixed in https://github.com/grpc/grpc-java/pull/5996 which was released in version 1.24


igorbernstein2:
After some discussion with marcal, we decided that we should tag views with project/instance/table ids. So will need to create my own views for grpc level

PR #11, opened 2019-10-21
chore: add and run synthtool

kolea2:
split up in incremental commits

PR #12, opened 2019-10-21
Fixup root pom.xml

igorbernstein2:
Main
* Update parent pom version along with the client

Properties:
* Add vertical space to properties to separate version from other variables
* alphabetize version props to make it easier to track
* only include primary dep version as a property
* add missing project.javadoc.protobufBaseURL property

Dependency Management:
* Make ordering consistent: Setions for BOMs, child modules, prod deps, tests. Each section is alphabetical
* move boms to top
* Child modules next
* Fix x-version-update markers
* Remove scopes from dependency management. It should only manage versions

Misc
* remove <ignoredUnusedDeclaredDependencies>objenesis, it's not used and causing warnings

### DIFF
--- a/google-cloud-bigtable/pom.xml
+++ b/google-cloud-bigtable/pom.xml
@@ -23,7 +23,7 @@
 
     <!-- Production dependencies -->
     <dependency>
-      <groupId>${project.groupId}</groupId>
+      <groupId>com.google.cloud</groupId>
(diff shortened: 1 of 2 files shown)


kolea2:
Let's hold off on this until #11 is merged. I had to make some other changes to get the build to pass

PR #22, opened 2019-10-22
chore: Clean up remaining files

kolea2:
- Move .repo-metadata.json, DEVELOPING.md, and README.md out of google-cloud-bigtable
- Update .repo-metadata.json with new repo name and release level
- Delete synth.py and synth.metadata from google-cloud-bigtable (newer version already exists at parent level)
- Add versions.txt


.repo-metadata.json
```diff
@@ -0,0 +1,13 @@
...
+  "name": "bigtable",
+  "name_pretty": "Cloud Bigtable",
+  "product_documentation": "https://cloud.google.com/bigtable",
+  "client_documentation": "https://googleapis.dev/java/java-bigtable/latest/index.html",
+  "issue_tracker": "https://issuetracker.google.com/savedsearches/559777",
+  "release_level": "ga",
+  "language": "java",
+  "repo": "googleapis/java-bigtable",
+  "repo_short": "java-bigtable",
+  "distribution_name": "google-cloud-bigtable",
```
kolea2: @chingor13 just want to confirm this is correct? The old file had `com.google.cloud:google-cloud-bigtable`

chingor13: It's not well defined how this field is meant to be used. For other languages, it's the value you add to `packages.json` or your `Gemfile`, but for Java, there's no singular value to put (some package managers split group/artifact/version). 

`groudId:artifactId:version` is most correct, but difficult to maintain. `groupId:artifactId` does seem more correct than just `artifactId`.

kolea2: ok, I'll switch it back to `groupId:artifactId`. Thanks!

PR #38, opened 2019-10-24
chore: Adding .gitignore to skip generated meta files & folders

rahulKQL:
Copied `.gitignore` content from the source(google-cloud-java) repository.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,42 @@
+# Packages
+dist
+bin
+var
+sdist
+target
+
+# Unit test / coverage reports
+.coverage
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,42 @@
+# Packages
```
kolea2: Let's also add `.gitignore` as per https://github.com/googleapis/google-cloud-java/blob/master/.gitignore#L1

kolea2:
Ignore my comment about .gitignore :) LGTM

PR #39, opened 2019-10-24
Added note about timestamp granularity in MutationApi interface.

rahulKQL:
For documentation about the timestamp format for Mutation#setCell APIs.


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/MutationApi.java
```diff
@@ -35,6 +35,10 @@
    * Adds a mutation which sets the value of the specified cell.
    *
    * <p>This is a convenience override that converts Strings to ByteStrings.
+   *
+   * <p>Note: The timestamp value could be -1 to use Bigtable server time. Otherwise, the user must
```
igorbernstein2: we should not be advertising server timestamps. They don't play well with retries

rahulKQL: Thanks for the review, I have reworded the JavaDoc. Please take a fresh look.

PR #42, opened 2019-10-24
chore(deps): update google-cloud-shared-config

igorbernstein2:
This should supercedes #15. 

The new shared config force checkstyle checks, unfortunately this doesn't play well with annotations and generated sources as maven will add the generated sources path to `$project.compileSourceRoots`, which checkstyle uses as the default value for `<sourceDirectories>`, this PR changes the value to src/main/java

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -364,6 +364,9 @@
                         <!-- Explicitly set the source directory to avoid running checkstyle on generated sources. -->
                         <sourceDirectory>src/main/java</sourceDirectory>
                       </sourceDirectories>
+                      <testSourceDirectories>
+                        <testSourceDirectory>src/test/java</testSourceDirectory>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -354,6 +354,19 @@
...
                 </plugin>
+
+                  <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-checkstyle-plugin</artifactId>
+                    <configuration>
+                      <sourceDirectories>
+                        <!-- TODO: pull this into google-cloud-shared-config -->
+                        <!-- Explicitly set the source directory to avoid running checkstyle on generated sources. -->
+                        <sourceDirectory>src/main/java</sourceDirectory>
```
kolea2: as discussed let's add for test directory

PR #44, opened 2019-10-25
chore: Temporarily move DirectPath tests back to the test endpoint.

igorbernstein2:
Unfortunately the prod DirectPath endpoint is not available yet. To unblock internal testing, update the test to target the test endpoint

### DIFF
--- a/google-cloud-bigtable/pom.xml
+++ b/google-cloud-bigtable/pom.xml
@@ -206,8 +206,11 @@
                   <systemPropertyVariables>
                     <bigtable.env>emulator</bigtable.env>
                   </systemPropertyVariables>
-                  <test>com.google.cloud.bigtable.data.v2.it.**</test>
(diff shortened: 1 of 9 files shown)


google-cloud-bigtable/pom.xml
```diff
@@ -232,10 +235,12 @@
                 </goals>
                 <configuration>
                   <systemPropertyVariables>
-                    <bigtable.env>prod</bigtable.env>
+                    <bigtable.env>cloud</bigtable.env>
```
kolea2: Just thought of this - do we need to update integration.cfg to include `bigtable-directpath-it`?

igorbernstein2: yes, but we need to whitelist the gcloud project first. This is the thing I mentioned we needed a separate kokoro pool for, but now that it's available in prod, we dont

PR #47, opened 2019-10-28
fix: fix maven test configs to make sure admin tests don't clobber data tests

igorbernstein2:


google-cloud-bigtable/pom.xml
```diff
@@ -206,8 +206,11 @@
                   <systemPropertyVariables>
                     <bigtable.env>emulator</bigtable.env>
                   </systemPropertyVariables>
-                  <test>com.google.cloud.bigtable.data.v2.it.**</test>
-                  <test>com.google.cloud.bigtable.admin.v2.it.**</test>
+                  <includes>
+                    <include>com.google.cloud.bigtable.data.v2.it.**</include>
```
kolea2: is it possible to replace this with a one-liner: `com.google.cloud.bigtable.**.v2.it.**`? Not super familiar with the syntax here but wanted to check!

igorbernstein2: Yep, updated. Thanks for the suggestion

PR #50, opened 2019-10-29
feat: add bom

### LINKED ISSUE #48: Add a BOM artifact
Please publish a google-cloud-bigtable-bom pom artifact that contains the synced versions of artifacts from this repository. We will consume them upstream in the google-cloud-bom as a single version rather than trying to keep 7+ artifact versions synced.

kolea2:
Fixes #48

### DIFF
--- a/google-cloud-bigtable-bom/pom.xml
+++ b/google-cloud-bigtable-bom/pom.xml
@@ -1,0 +1,116 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.google.cloud</groupId>
+    <artifactId>google-cloud-bigtable-bom</artifactId>
+    <version>1.6.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
+    <packaging>pom</packaging>
+    <parent>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-shared-config</artifactId>
+        <version>0.2.1</version>
+    </parent>
+
+    <name>Google Cloud Bigtable BOM</name>
+    <url>https://github.com/googleapis/java-bigtable</url>
+    <description>
+        BOM for Google Cloud Bigtable
+    </description>
+
+    <organization>
+        <name>Google LLC</name>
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable-bom/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.google.cloud</groupId>
+    <artifactId>google-cloud-bigtable-bom</artifactId>
+    <version>1.6.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
+    <packaging>pom</packaging>
+    <parent>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-shared-config</artifactId>
+        <version>0.2.1</version>
+    </parent>
```
igorbernstein2: I'm not sure if this is right.  This bom is meant to be imported by end users, but the shared config includes plugin configs  that are meant to be used by client developers

google-cloud-bigtable-bom/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+        </developer>
+        <developer>
+            <id>igorberstein</id>
+            <name>Igor Bernstein</name>
+            <email>igorbernstein@google.com</email>
+            <organization>Google LLC</organization>
+            <roles>
+                <role>Developer</role>
+            </roles>
+        </developer>
```
igorbernstein2: I think we should add Kristen here as well

google-cloud-bigtable-bom/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
```
igorbernstein2: newline please

google-cloud-bigtable-bom/pom.xml
chingor13: Unless end users are using this artifact as their parent, they won't inherit the build rules in `google-cloud-shared-config`. We use that parent for all the release configuration (no compile/test/runtime dependency management in there)

igorbernstein2: Looks like other projects are using this pattern:
https://github.com/jersey/jersey/blob/faa809da43538ce31076b50f969b4bd64caa5ac9/bom/pom.xml#L48-L53
https://github.com/sonatype/oss-parents/blob/5cc24e62cae0fc8a1a6a917c581574b71dec476f/jvnet-parent/pom.xml

PR #58, opened 2019-10-31
add test that injects failures in Directpath path while test is running

vnorigoog:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for context and/or discussion)

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,9 +335,6 @@ Java 7 or above is required for using this client.
 
 This library follows [Semantic Versioning](http://semver.org/).
 
-It is currently in major version zero (`0.y.z`), which means that anything may
-change at any time and the public API should not be considered stable.
-
 ## Contributing
 
 Contributions to this library are always welcome and highly encouraged.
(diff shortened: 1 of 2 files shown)


vnorigoog:
@igorbernstein2 : This change is mostly meant for running the tests on a VM that is created explicitly for testing purposes (such as, presubmit tests).
Your point is very valid - about machine being messed up if it is run on your workstation.

2 things I would say are interesting about this approach are
1. this mechanism of injecting failures is portable to any language
2. injection of failures happens outside of grpc - at the machine level. we don't have to worry about the internals of netty and such.

If you feel *strongly* that this mechanism of failures injection is *not* needed or necessary, please feel free to say "we shouldn't do this" or "we don't need it". I will be happy to discard to this pull request.

vnorigoog:
btw, this is still work in progress and I am not yet ready to send for review.

If this PR is deemed useful, then I will request review sometime next week, after testing it out completely.

igorbernstein2:
I think you have a valid point about the approach being applicable cross language. However to get benefit from that approach, it should be implemented outside of the client. I’m working on a proposal for cross language client test suite and I think that this approach would be better suited there. 

go/cbt-client-conformance-tests

I definitely would like to avoid corrupting a developer’s workstation after running mvn verify

vnorigoog:
+1 to your proposal
"I’m working on a proposal for cross language client test suite and I think that this approach would be better suited there."

I am going to abandon this PR.
thanks for reviewing.

vnorigoog:
deleted

PR #79, opened 2019-11-11
chore: Ignore gapic settings for all streaming RPCs

igorbernstein2:
This is to prepare for upcoming changes that will move simple retries to gRPC.
When this happens we want to retain retry settings at the client level, so they are copied into the manual settings wrapper.

See cl/277568516 for more details

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
@@ -117,7 +117,7 @@ public class EnhancedBigtableStubSettings extends StubSettings<EnhancedBigtableS
           .setJittered(true)
           .setInitialRpcTimeout(Duration.ofMinutes(5))
           .setRpcTimeoutMultiplier(2.0)
-          .setInitialRpcTimeout(Duration.ofMinutes(5))
+          .setMaxRpcTimeout(Duration.ofMinutes(5))
           .setTotalTimeout(Duration.ofHours(12))
           .build();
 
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -102,6 +102,25 @@
...
 
+  // Allow retrying ABORTED statuses. These will be returned by the server when the client is
+  // too slow to read the rows. This makes sense for the java client because retries happen
+  // after the row merging logic. Which means that the retry will not be invoked until the
+  // current buffered chunks are consumed.
+  private static final Set<Code> READ_ROWS_RETRY_CODES =
+      ImmutableSet.<Code>builder().addAll(IDEMPOTENT_RETRY_CODES).add(Code.ABORTED).build();
+
+  private static final RetrySettings READ_ROWS_RETRY_SETTINGS =
+      RetrySettings.newBuilder()
```
kolea2: where do these defaults come from?

kolea2: (should they be extracted from `BigtableStubSettings`?  I notice they're slightly different than read_rows_params https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableStubSettings.java#L260-L270)

igorbernstein2: Hmm, thats quite odd....the generated code doesn't match the yaml:
https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable_gapic.yaml#L56-L63

The yaml one is correct. At some point we should investigate why the two diverged

PR #82, opened 2019-11-11
chore: fix integration test zones to work across all environments

igorbernstein2:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for context and/or discussion)


rahulKQL:
Considering this file is being ignored by CodeCov, So merging this change now.

PR #83, opened 2019-11-12
chore: fix tests

igorbernstein2:
* Fix bug in the instance admin IT. This was unnoticed because our presubmits target a development instance (I just changed that)
* Move auto-value back as a provided dep in the main dependency list. This caused issues in intellij when annotation processing was disabled
* Disable concurrent unit tests, it makes the autogen tests flaky

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableInstanceAdminClientIT.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableInstanceAdminClientIT.java
@@ -42,7 +42,8 @@ import org.threeten.bp.Instant;
 
 public class BigtableInstanceAdminClientIT {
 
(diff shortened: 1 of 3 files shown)


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableInstanceAdminClientIT.java
```diff
@@ -143,29 +146,25 @@ public void instanceCreationDeletionTest() {
     }
   }
 
-  @Test
-  public void clusterCreationDeletionTest() {
-    Instance currentInstance = client.getInstance(instanceId);
-    assume()
```
kolea2: is this no longer true?

igorbernstein2: it uses a brand new instance from instanceAndClusterCreationDeletionTest, which is always prod

PR #91, opened 2019-11-15
chore: update Error message for bad bigtable.env variable

sid-dinesh94:
At some point, the allowed settings were changed from "prod" and "emulator" to "cloud" and "emulator".
This change updates the error message associated to the illegal argument exception.

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/TestEnvRule.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/TestEnvRule.java
@@ -103,7 +103,7 @@ public class TestEnvRule implements TestRule {
       default:
         throw new IllegalArgumentException(
             String.format(
-                "Unknown env: %s. Please set the system property %s to either 'emulator' or 'prod'.",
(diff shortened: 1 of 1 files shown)


igorbernstein2:
Yikes! Sorry about that and thank you for the fix!

However I was wondering if you are modifying that property directly? It's meant to be controlled via a maven profile:
https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable/pom.xml#L260

ie: `mvn verify -P bigtable-prod-it`

igorbernstein2:
Hope you don't mind, I amended your commit to comply with conventional commit messages

sid-dinesh94:
Thanks for the work in checking this in!
We have a need to create the test tables within a VPC Service Controlled security perimeter. This requires us to customize the environment variables.

igorbernstein2:
May I suggest to use a combination?

‘’’bash

mvn verify -Pbigtable-prod-it -Dbigtable.project=... -Dbigtable.instance=... -Dbigtable.table=....
‘’’

The bigtable.env property wasn’t meant to be set directly

sid-dinesh94:
Perfect, thank you for the advice. I will modify the invocation accordingly.

PR #92, opened 2019-11-18
Added `FILTERS.timestamp().exact()` for better access & read ability.

### LINKED ISSUE #65: Add syntactic sugar for exact timestamp lookups
`FILTERS.timestamp().range().startClosed(x).endClosed(x)` is a bit verbose
`FILTERS.timestamp().exact(x)` would be preferable

rahulKQL:
Fixes #65

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
@@ -351,6 +351,21 @@ public class FiltersTest {
   }
 
   @Test
+  public void timestampAtExactTime() {
+    RowFilter actualFilter = FILTERS.timestamp().exact(20_000L).toProto();
+
+    RowFilter expectedFilter =
+        RowFilter.newBuilder()
+            .setTimestampRangeFilter(
+                TimestampRange.newBuilder()
+                    .setStartTimestampMicros(20_000L)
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
```diff
@@ -350,6 +350,21 @@ public void timestampRange() {
...
+  @Test
+  public void timestampAtExactTime() {
+    RowFilter actualFilter = FILTERS.timestamp().exact(20_000L).toProto();
+
+    RowFilter expectedFilter =
+        RowFilter.newBuilder()
+            .setTimestampRangeFilter(
+                TimestampRange.newBuilder()
+                    .setStartTimestampMicros(20_000L)
+                    .setEndTimestampMicros(20_000L + 1))
```
kolea2: why `+ 1` here?

rahulKQL: `TimestampRange#endTimestampMicros` denotes upper bound for timestamp range. If we pass the same value then no rows will return.

igorbernstein2: The end range cap is exclusive, so for a single number, the range is [x, x+1)

PR #99, opened 2019-11-22
feat: introducing bulk read API through Batcher

### LINKED ISSUE #7: Add bulk read api
bigtable-client-core currently has a feature to use ReadRows for bulk reads. I think we can easily  port over this functionality using the Batcher api similar to how bulk mutations work:

* Add `Batcher<ByteString, Row> newBulkReadBatcher(tableId, filter)` to the client and stub surface
  * Have it return a instance BatcherImpl instance that wraps the ReadRowsCallable
* Add a BulkReadRowsRequestBuilder that wraps a Query and adds ByteString row keys
* Add a ReadRowsBatchingDescriptor that can map results back to entry futures
* Expose batching read settings as bulkReadRowsSettings on the stub

Things to consider:
* Need to be careful with tracing to avoid creating confusing spans
* Need to make sure to expose the ability to set a filter
* Need to be careful with UnaryCallableSettings to make sure that bulk & streaming settings dont conflict

rahulKQL:
Fixes #7 

This change introduces BulkReadAPI on BigtableDataClient. This operation
accepts row keys in a batch mode and behind the scene split them based
on configurable values.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -26,6 +26,7 @@ import com.google.api.gax.rpc.ServerStreamingCallable;
 import com.google.api.gax.rpc.UnaryCallable;
 import com.google.cloud.bigtable.data.v2.models.BulkMutation;
 import com.google.cloud.bigtable.data.v2.models.ConditionalRowMutation;
+import com.google.cloud.bigtable.data.v2.models.Filters;
 import com.google.cloud.bigtable.data.v2.models.Filters.Filter;
 import com.google.cloud.bigtable.data.v2.models.KeyOffset;
 import com.google.cloud.bigtable.data.v2.models.Query;
@@ -934,6 +935,83 @@ public class BigtableDataClient implements AutoCloseable {
   @BetaApi("This surface is likely to change as the batching surface evolves.")
   public Batcher<RowMutationEntry, Void> newBulkMutationBatcher(@Nonnull String tableId) {
     return stub.newMutateRowsBatcher(tableId);
+  }
+
+  /**
+   * Reads rows for given tableId in a batch. If the row does not exist, the value will be null.
+   * This operation should be called with in a single thread.
+   *
+   * <p>Sample Code:
+   *
+   * <pre>{@code
+   * try (BigtableDataClient bigtableDataClient = BigtableDataClient.create("[PROJECT]", "[INSTANCE]")) {
+   *   List<ApiFuture<Row>> rows = new ArrayList<>();
+   *
+   *   try (Batcher<ByteString, Row> batcher = bigtableDataClient.newBulkReadRowsBatcher("[TABLE]")) {
+   *     for (String someValue : someCollection) {
+   *       ApiFuture<Row> rowFuture =
+   *           batcher.add(ByteString.copyFromUtf8("[ROW KEY]"));
+   *       rows.add(rowFuture);
+   *     }
+   *
+   *     // Sends collected elements for batching asynchronously.
+   *     batcher.sendOutstanding();
+   *
+   *     // Blocks until all pending row keys are fetched from Bigtable.
+   *     batcher.flush();
+   *   }
+   *   // Before `batcher` is closed, all remaining(If any) rows are fetched.
+   *
+   *   List<Row> actualRows = ApiFutures.allAsList(rows).get();
+   * }
+   * }</pre>
+   */
+  public Batcher<ByteString, Row> newBulkReadRowsBatcher(String tableId) {
+    return newBulkReadRowsBatcher(tableId, null);
+  }
+
+  /**
+   * Reads rows for given tableId and filter criteria in a batch. If the row does not exist, the
+   * value will be null. This operation should be called with in a single thread.
+   *
+   * <p>Sample Code:
+   *
+   * <pre>{@code
+   * try (BigtableDataClient bigtableDataClient = BigtableDataClient.create("[PROJECT]", "[INSTANCE]")) {
+   *
+   *  // Build the filter expression
+   *  Filter filter = FILTERS.chain()
+   *         .filter(FILTERS.key().regex("prefix.*"))
+   *         .filter(FILTERS.limit().cellsPerRow(10));
+   *
+   *   List<ApiFuture<Row>> rows = new ArrayList<>();
+   *
+   *   try (Batcher<ByteString, Row> batcher = bigtableDataClient.newBulkReadRowsBatcher("[TABLE]", filter)) {
+   *     for (String someValue : someCollection) {
+   *       ApiFuture<Row> rowFuture =
+   *           batcher.add(ByteString.copyFromUtf8("[ROW KEY]"));
+   *       rows.add(rowFuture);
+   *     }
+   *
+   *     // Sends collected elements for batching asynchronously.
+   *     batcher.sendOutstanding();
+   *
+   *     // Blocks until all pending row keys are fetched from Bigtable.
+   *     batcher.flush();
+   *   }
+   *   // Before `batcher` is closed, all remaining(If any) rows are fetched.
+   *
+   *   List<Row> actualRows = ApiFutures.allAsList(rows).get();
+   * }
+   * }</pre>
+   */
+  public Batcher<ByteString, Row> newBulkReadRowsBatcher(
+      String tableId, @Nullable Filters.Filter filter) {
+    Query query = Query.create(tableId);
+    if (filter != null) {
+      query = query.filter(filter);
+    }
+    return stub.newBulkReadRowsBatcher(query);
   }
 
   /**
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -954,13 +954,14 @@ public class BigtableDataClient implements AutoCloseable {
    *       rows.add(rowFuture);
    *     }
    *
-   *     // Sends collected elements for batching asynchronously.
+   *     // [Optional] Sends collected elements for batching asynchronously.
    *     batcher.sendOutstanding();
    *
-   *     // Blocks until all pending row keys are fetched from Bigtable.
+   *     // [Optional] Invokes sendOutstanding() and awaits until all pending entries are resolved.
    *     batcher.flush();
    *   }
-   *   // Before `batcher` is closed, all remaining(If any) rows are fetched.
+   *   // batcher.close() invokes `flush()` which intern invoke `sendOutstanding()` with await for
+   *   pending batches until its resolved.
    *
    *   List<Row> actualRows = ApiFutures.allAsList(rows).get();
    * }
@@ -993,13 +994,14 @@ public class BigtableDataClient implements AutoCloseable {
    *       rows.add(rowFuture);
    *     }
    *
-   *     // Sends collected elements for batching asynchronously.
+   *     // [Optional] Sends collected elements for batching asynchronously.
    *     batcher.sendOutstanding();
    *
-   *     // Blocks until all pending row keys are fetched from Bigtable.
+   *     // [Optional] Invokes sendOutstanding() and awaits until all pending entries are resolved.
    *     batcher.flush();
    *   }
-   *   // Before `batcher` is closed, all remaining(If any) rows are fetched.
+   *   // batcher.close() invokes `flush()` which intern invoke `sendOutstanding()` with await for
+   *   pending batches until its resolved.
    *
    *   List<Row> actualRows = ApiFutures.allAsList(rows).get();
    * }
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -960,7 +960,7 @@ public class BigtableDataClient implements AutoCloseable {
    *     // [Optional] Invokes sendOutstanding() and awaits until all pending entries are resolved.
    *     batcher.flush();
    *   }
-   *   // batcher.close() invokes `flush()` which intern invoke `sendOutstanding()` with await for
+   *   // batcher.close() invokes `flush()` which will in turn invoke `sendOutstanding()` with await for
    *   pending batches until its resolved.
    *
    *   List<Row> actualRows = ApiFutures.allAsList(rows).get();
@@ -1000,7 +1000,7 @@ public class BigtableDataClient implements AutoCloseable {
    *     // [Optional] Invokes sendOutstanding() and awaits until all pending entries are resolved.
    *     batcher.flush();
    *   }
-   *   // batcher.close() invokes `flush()` which intern invoke `sendOutstanding()` with await for
+   *   // batcher.close() invokes `flush()` which will in turn invoke `sendOutstanding()` with await for
    *   pending batches until its resolved.
    *
    *   List<Row> actualRows = ApiFutures.allAsList(rows).get();
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
@@ -38,6 +38,7 @@ import com.google.cloud.bigtable.data.v2.internal.RequestContext;
 import com.google.cloud.bigtable.data.v2.models.BulkMutation;
 import com.google.cloud.bigtable.data.v2.models.ConditionalRowMutation;
 import com.google.cloud.bigtable.data.v2.models.DefaultRowAdapter;
+import com.google.cloud.bigtable.data.v2.models.Filters;
 import com.google.cloud.bigtable.data.v2.models.KeyOffset;
 import com.google.cloud.bigtable.data.v2.models.Query;
 import com.google.cloud.bigtable.data.v2.models.ReadModifyWriteRow;
@@ -52,12 +53,14 @@ import com.google.cloud.bigtable.data.v2.stub.mutaterows.BulkMutateRowsUserFacin
 import com.google.cloud.bigtable.data.v2.stub.mutaterows.MutateRowsBatchingDescriptor;
 import com.google.cloud.bigtable.data.v2.stub.mutaterows.MutateRowsRetryingCallable;
 import com.google.cloud.bigtable.data.v2.stub.readrows.FilterMarkerRowsCallable;
+import com.google.cloud.bigtable.data.v2.stub.readrows.ReadRowsBatchingDescriptor;
 import com.google.cloud.bigtable.data.v2.stub.readrows.ReadRowsResumptionStrategy;
 import com.google.cloud.bigtable.data.v2.stub.readrows.ReadRowsRetryCompletedCallable;
 import com.google.cloud.bigtable.data.v2.stub.readrows.ReadRowsUserCallable;
 import com.google.cloud.bigtable.data.v2.stub.readrows.RowMergingCallable;
 import com.google.cloud.bigtable.gaxx.retrying.ApiResultRetryAlgorithm;
 import com.google.cloud.bigtable.gaxx.tracing.WrappedTracerFactory;
+import com.google.protobuf.ByteString;
 import io.opencensus.stats.Stats;
 import io.opencensus.stats.StatsRecorder;
 import io.opencensus.tags.Tagger;
@@ -65,6 +68,7 @@ import io.opencensus.tags.Tags;
 import java.io.IOException;
 import java.util.List;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.threeten.bp.Duration;
 
 /**
@@ -381,9 +385,8 @@ public class EnhancedBigtableStub implements AutoCloseable {
   }
 
   /**
-   * Creates a {@link com.google.api.gax.batching.BatcherImpl} to handle {@link
-   * MutateRowsRequest.Entry} mutations. This is meant to be used for automatic batching with flow
-   * control.
+   * Creates a {@link BatcherImpl} to handle {@link MutateRowsRequest.Entry} mutations. This is
+   * meant to be used for automatic batching with flow control.
    *
    * <ul>
    *   <li>Uses {@link MutateRowsBatchingDescriptor} to spool the {@link RowMutationEntry} mutations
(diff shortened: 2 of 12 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsBatchingDescriptor.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Override
+  public BatchingRequestBuilder<ByteString, Query> newRequestBuilder(Query query) {
+    return new BulkReadRowsRequestBuilder(query);
+  }
+
+  @Override
+  public void splitResponse(List<Row> rows, List<SettableApiFuture<Row>> futureResults) {
+    // TODO: verify if this check
+    Preconditions.checkState(rows.size() == futureResults.size());
```
igorbernstein2: This needs to be a bit more careful:
- you need to consider cases when keys don’t exist
- when there are duplicate keys

Which means that we need to able to match the original requested key to the row key. 

I need to take a closer look, but it seems like this will require changes to the gax BatchingDescriptor api...it needs to access the original request entries.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -503,6 +511,25 @@ private Builder() {
...
+      bulkReadRowsSettings =
+          BigtableBulkReadRowsSettings.newBuilder(new ReadRowsBatchingDescriptor())
+              .setRetryableCodes(IDEMPOTENT_RETRY_CODES)
+              .setRetrySettings(MUTATE_ROWS_RETRY_SETTINGS)
+              .setBatchingSettings(
+                  // TODO: Need to finalize initial thresholds for Read
+                  BatchingSettings.newBuilder()
+                      .setIsEnabled(true)
+                      .setElementCountThreshold(100L)
+                      .setRequestByteThreshold(20L * 1024 * 1024)
```
igorbernstein2: I don't think a byte threshold is necessary in this case

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -503,6 +511,25 @@ private Builder() {
...
+                      .setDelayThreshold(Duration.ofSeconds(1))
+                      .setFlowControlSettings(
+                          FlowControlSettings.newBuilder()
+                              .setLimitExceededBehavior(LimitExceededBehavior.Block)
+                              .setMaxOutstandingRequestBytes(100L * 1024 * 1024)
```
igorbernstein2: Same as above, dont really care about outstanding bytes

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -409,6 +411,18 @@ public static EnhancedBigtableStub create(EnhancedBigtableStubSettings settings)
         clientContext.getExecutor());
   }
 
+  // TODO: javadoc
+  public Batcher<ByteString, Row> newBulkReadRowsBatcher(
+      @Nonnull String tableId, @Nonnull Filters.Filter filter) {
```
igorbernstein2: There should be a variant w/o a filter as well

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsBatchingDescriptor.java
```diff
@@ -0,0 +1,77 @@
...
+
+    for (int rowCount = 0; rowCount < rows.size(); rowCount++) {
+      futureResults.get(rowCount).set(rows.get(rowCount));
+    }
+  }
+
+  @Override
+  public void splitException(Throwable throwable, List<SettableApiFuture<Row>> list) {
+    // TODO: do we have to consider specific exception?
```
igorbernstein2: I dont think so

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsBatchingDescriptor.java
igorbernstein2: Also this needs unit tests to handle those cases

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +563,29 @@ private Builder() {
...
+              .setBatchingSettings(
+                  BatchingSettings.newBuilder()
+                      .setIsEnabled(true)
+                      .setElementCountThreshold(100L)
+                      .setRequestByteThreshold(2L * 1024L * 1024)
+                      .setDelayThreshold(Duration.ofSeconds(1))
+                      .setFlowControlSettings(
+                          FlowControlSettings.newBuilder()
+                              .setLimitExceededBehavior(LimitExceededBehavior.Block)
+                              .setMaxOutstandingElementCount(1_000L)
```
igorbernstein2: with 100 elements/batch this means that we can't have more than 10 outstanding streams. Maybe this should be 10 * 100 * numChannels? So that there can be 10 outstanding requests per channel?

rahulKQL: > Maybe this should be 10 * 100 * numChannels?

Sounds good except doubt for one scenario: What if enduser has say 32 channels (n1-stnd-8 machine), as each batch is unaryCallable(unlike readRows streaming) and we can not apply flow control based on actual row size. 
_Do you think it should be okay with these cases?_

We need numChannels which come from CPU so Instead of numChannels, Would `20 * 100 * numOfCPUs` be ok?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +563,29 @@ private Builder() {
...
+                      .getRetrySettings()
+                      .toBuilder()
+                      .setMaxRpcTimeout(Duration.ofSeconds(300))
+                      .setTotalTimeout(Duration.ofMinutes(20))
+                      .build())
+              .setBatchingSettings(
+                  BatchingSettings.newBuilder()
+                      .setIsEnabled(true)
+                      .setElementCountThreshold(100L)
+                      .setRequestByteThreshold(2L * 1024L * 1024)
```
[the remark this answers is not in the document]
rahulKQL: Thanks alot for the details, I will update this number.

And for filter size, I would raise it in a follow-up PR.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
igorbernstein2: I think we should chat tomorrow. I don't quite understand what you mean here.

Minor correction: the default settings are 2x the cpus, so -8 will have 16 channels. I think making it based on channels is more semantic

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -936,6 +937,83 @@ public void bulkMutateRows(BulkMutation mutation) {
...
+   *       ApiFuture<Row> rowFuture =
+   *           batcher.add(ByteString.copyFromUtf8("[ROW KEY]"));
+   *       rows.add(rowFuture);
+   *     }
+   *
+   *     // Sends collected elements for batching asynchronously.
+   *     batcher.sendOutstanding();
+   *
+   *     // Blocks until all pending row keys are fetched from Bigtable.
+   *     batcher.flush();
```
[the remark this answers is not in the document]
rahulKQL: I have updated samples, Please have a look and let me if that looks fine.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -936,6 +937,85 @@ public void bulkMutateRows(BulkMutation mutation) {
...
+   *       rows.add(rowFuture);
+   *     }
+   *
+   *     // [Optional] Sends collected elements for batching asynchronously.
+   *     batcher.sendOutstanding();
+   *
+   *     // [Optional] Invokes sendOutstanding() and awaits until all pending entries are resolved.
+   *     batcher.flush();
+   *   }
+   *   // batcher.close() invokes `flush()` which intern invoke `sendOutstanding()` with await for
```
igorbernstein2: s/in tern/will in turn/

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -936,6 +937,85 @@ public void bulkMutateRows(BulkMutation mutation) {
...
```
igorbernstein2: same as above

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -384,6 +388,49 @@ public BigtableBatchingCallSettings bulkMutateRowsSettings() {
...
+   * </ul>
+   *
+   * <p>On breach of certain triggers, the operation initiates processing of accumulated request for
+   * which the default settings are:
+   *
+   * <ul>
+   *   <li>When the {@link BatchingSettings.Builder#setElementCountThreshold request count} reaches
+   *       100.
+   *   <li>When accumulated {@link BatchingSettings.Builder#setRequestByteThreshold request size}
+   *       reaches to 20MB.
```
igorbernstein2: This is stale

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -384,6 +388,49 @@ public BigtableBatchingCallSettings bulkMutateRowsSettings() {
...
+   *   <li>When an {@link BatchingSettings.Builder#setDelayThreshold interval of} 1 second passes
+   *       after batching initialization or last processed batch.
+   * </ul>
+   *
+   * <p>When the pending {@link FlowControlSettings.Builder#setMaxOutstandingElementCount request
+   * count} reaches a default of 1000 or their {@link
+   * FlowControlSettings.Builder#setMaxOutstandingRequestBytes accumulated size} reaches default
+   * value of 100MB, then this operation will by default be {@link
```
igorbernstein2: This also looks stale

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -384,6 +388,49 @@ public BigtableBatchingCallSettings bulkMutateRowsSettings() {
...
+   *   <li>When the {@link BatchingSettings.Builder#setElementCountThreshold request count} reaches
+   *       100.
+   *   <li>When accumulated {@link BatchingSettings.Builder#setRequestByteThreshold request size}
+   *       reaches to 20MB.
+   *   <li>When an {@link BatchingSettings.Builder#setDelayThreshold interval of} 1 second passes
+   *       after batching initialization or last processed batch.
+   * </ul>
+   *
+   * <p>When the pending {@link FlowControlSettings.Builder#setMaxOutstandingElementCount request
+   * count} reaches a default of 1000 or their {@link
```
igorbernstein2: Please make this clear that this refers to number of outstanding keys

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +563,30 @@ private Builder() {
                               .build())
                       .build());
 
+      long numOfChannels = 2 * Runtime.getRuntime().availableProcessors();
```
igorbernstein2: I think you should extract this to a static package protected method and use it here and in defaultGrpcTransportProviderBuilder

```
static int getDefaultChannelPoolSize() {
   return 2 * Runtime.getRuntime().availableProcessors();
}
public static defaultGrpcTransportProviderBuilder() {
return BigtableStubSettings.defaultGrpcTransportProviderBuilder()
  .setPoolSize(getDefaultChannelPoolSize())
  .setMaxInboundMessageSize(268435456);
}
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +563,30 @@ private Builder() {
...
+      bulkReadRowsSettings =
+          BigtableBulkReadRowsCallSettings.newBuilder(new ReadRowsBatchingDescriptor())
+              .setRetryableCodes(readRowSettings.getRetryableCodes())
+              .setRetrySettings(
+                  readRowSettings
+                      .getRetrySettings()
+                      .toBuilder()
+                      .setMaxRpcTimeout(Duration.ofSeconds(300))
```
igorbernstein2: Duration.ofMinutes(5)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +563,30 @@ private Builder() {
                               .build())
                       .build());
 
+      long numOfChannels = 2 * Runtime.getRuntime().availableProcessors();
+      bulkReadRowsSettings =
+          BigtableBulkReadRowsCallSettings.newBuilder(new ReadRowsBatchingDescriptor())
+              .setRetryableCodes(readRowSettings.getRetryableCodes())
```
igorbernstein2: lease add a precondition.checkstate to EnhancedBigtableStubSettings(Builder builder) that makes sure that the codes here match `.readRowSettings.getRetryableCodes()`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +563,30 @@ private Builder() {
...
+              .setBatchingSettings(
+                  BatchingSettings.newBuilder()
+                      .setIsEnabled(true)
+                      .setElementCountThreshold(100L)
+                      .setRequestByteThreshold(400L * 1024L)
+                      .setDelayThreshold(Duration.ofSeconds(1))
+                      .setFlowControlSettings(
+                          FlowControlSettings.newBuilder()
+                              .setLimitExceededBehavior(LimitExceededBehavior.Block)
+                              .setMaxOutstandingElementCount(10L * 100L * numOfChannels)
```
igorbernstein2: please extract 100 to a local variable to make it obvious that its shared with element count threshold

long maxElementPerBatch = 100L;

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/QueryTest.java
```diff
@@ -286,4 +286,29 @@ public void testEquality() {
...
+
+    assertThat(query.toProto(requestContext)).isEqualTo(request);
+
+    assertThat(Query.create(TABLE_ID).limit(5).toProto(requestContext))
+        .isNotEqualTo(
+            ReadRowsRequest.newBuilder()
+                .setTableName(NameUtil.formatTableName(PROJECT_ID, INSTANCE_ID, TABLE_ID))
+                .setAppProfileId(APP_PROFILE_ID)
+                .setRowsLimit(10)
+                .build());
```
igorbernstein2: What are you testing here?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/BulkReadIT.java
```diff
@@ -0,0 +1,98 @@
...
+                      ByteString.copyFromUtf8("value-" + i)))));
+    }
+    client.bulkMutateRows(bulkMutation);
+
+    List<ApiFuture<Row>> rowFutures = new ArrayList<>();
+    try (Batcher<ByteString, Row> batcher =
+        client.newBulkReadRowsBatcher(testEnvRule.env().getTableId())) {
+
+      for (int rowCount = 0; rowCount < numRows; rowCount++) {
+        rowFutures.add(batcher.add(ByteString.copyFromUtf8(rowPrefix + "-" + rowCount)));
```
igorbernstein2: Please split this up into 3 lines to make it easier to read

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
rahulKQL: Have updated JavaDoc to mention that.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/QueryTest.java
rahulKQL: Thanks for pointing it out, Have updated this with valid test case.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -224,10 +234,14 @@ public boolean isRefreshingChannel() {
   public static InstantiatingGrpcChannelProvider.Builder defaultGrpcTransportProviderBuilder() {
     return BigtableStubSettings.defaultGrpcTransportProviderBuilder()
         // TODO: tune channels
```
igorbernstein2: todo should move to getDefaultChannelPoolSize

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +571,25 @@ private Builder() {
...
                       .build());
 
+      long maxElementPerBatch = 100L;
+      bulkReadRowsSettings =
+          BigtableBulkReadRowsCallSettings.newBuilder(new ReadRowsBatchingDescriptor())
+              .setRetryableCodes(readRowsSettings.getRetryableCodes())
+              .setRetrySettings(IDEMPOTENT_RETRY_SETTINGS)
+              .setBatchingSettings(
+                  BatchingSettings.newBuilder()
+                      .setIsEnabled(true)
```
kolea2: Is this required? Think this may be the default.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +571,25 @@ private Builder() {
...
+                      .setElementCountThreshold(maxElementPerBatch)
+                      .setRequestByteThreshold(400L * 1024L)
```
kolea2: can we make this a variable?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -515,6 +571,25 @@ private Builder() {
...
+                      .setDelayThreshold(Duration.ofSeconds(1))
+                      .setFlowControlSettings(
+                          FlowControlSettings.newBuilder()
+                              .setLimitExceededBehavior(LimitExceededBehavior.Block)
+                              .setMaxOutstandingElementCount(
+                                  10L * maxElementPerBatch * getDefaultChannelPoolSize())
```
kolea2: variable?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/BulkReadIT.java
```diff
@@ -0,0 +1,105 @@
...
+      // non-existent row key
+      rowFutures.add(batcher.add(ByteString.copyFromUtf8(UUID.randomUUID().toString())));
+
+      // duplicate row key
+      rowFutures.add(batcher.add(ByteString.copyFromUtf8(rowPrefix + "-" + 0)));
+      rowFutures.add(batcher.add(ByteString.copyFromUtf8(rowPrefix + "-" + 0)));
+
+      batcher.flush();
+      actualRows = ApiFutures.allAsList(rowFutures).get();
+      assertThat(actualRows.get(0)).isEqualTo(null);
```
kolea2: I think you can do `assertThat(actualRows.get(0)).isNull()`

kolea2:
Added a few small comments but LGTM overall!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
rahulKQL: BatcherV2 does not use this value at all, it was left for backward compatibility.

I had added this for semantic purpose, I would remove it from here.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
rahulKQL: We could have a variable here. Although as in this format, it is similar to the existing `bulkMutateRowsSettings` configuration
https://github.com/googleapis/java-bigtable/blob/abbc6c8c0ffbe1dcf88c14b1e70fe3150c61c582/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java#L500-L516

We might need to change that to have common formatting in separate PR

PR #102, opened 2019-11-26
deps: move auto-value to annotation processor path

chingor13:
This removes `auto-value` from the dependencies - `auto-value` is only needed at compile time. `auto-value-annotations` remains as a dependency.

### DIFF
--- a/google-cloud-bigtable/pom.xml
+++ b/google-cloud-bigtable/pom.xml
@@ -64,11 +64,6 @@
     </dependency>
     <dependency>
       <groupId>com.google.auto.value</groupId>
-      <artifactId>auto-value</artifactId>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
-      <groupId>com.google.auto.value</groupId>
       <artifactId>auto-value-annotations</artifactId>
     </dependency>
     <dependency>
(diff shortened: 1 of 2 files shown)


kolea2:
Hi @chingor13! We originally had it this way. We ended up reverting it here: https://github.com/googleapis/java-bigtable/pull/83, because running tests in IntelliJ wasn't working. It was having trouble compiling without going into the annotation processing settings and making changes there.

google-cloud-bigtable/pom.xml
```diff
@@ -376,6 +371,24 @@
...
+          <encoding>UTF-8</encoding>
+          <compilerArgument>-Xlint:unchecked</compilerArgument>
+          <annotationProcessorPaths>
+            <path>
+              <groupId>com.google.auto.value</groupId>
+              <artifactId>auto-value</artifactId>
+              <version>${autovalue.version}</version>
+            </path>
+          </annotationProcessorPaths>
+        </configuration>
```
igorbernstein2: This causes build issues in IntelliJ. See the attached screenshot
Screenshot from 2019-12-02 12-13-20

igorbernstein2: Actually nevermind, it fails with annotation processing enabled as well

igorbernstein2: It seems like intellij is failing to resolve transitive deps for `annotationProcessorPaths`, explicitly adding auto-service alongside auto-value fixes the intellij issue:

```
<plugin>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>3.5.1</version>
        <configuration>
          <source>1.7</source>
          <target>1.7</target>
          <encoding>UTF-8</encoding>
          <compilerArgument>-Xlint:unchecked</compilerArgument>
          <annotationProcessorPaths>
            <path>
              <groupId>com.google.auto.value</groupId>
              <artifactId>auto-value</artifactId>
              <version>${autovalue.version}</version>
            </path>
            <path>
              <groupId>com.google.auto.service</groupId>
              <artifactId>auto-service</artifactId>
              <version>1.0-rc6</version>
            </path>
          </annotationProcessorPaths>
        </configuration>
      </plugin>
```

But we would have to manually keep the 2 versions in sync. I don't think its worth the added maintenance cost

igorbernstein2:
I'm going to close this, please reopen if you disagree

PR #112, opened 2019-12-09
feat: add BigtableDataClientFactory to create lightweight data clients

igorbernstein2:
The new factory allows users to construct a single heavy factory object that can create many lightweight clients.
This is meant to be used in situations when a single application needs to access multiple instances or use different application profiles

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClientFactory.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClientFactory.java
@@ -61,6 +61,8 @@ import javax.annotation.Nonnull;
  * otherInstanceClient.close();
  *
  * clientFactory.close();
+ *
+ * <p>Please note that this is an experimental feature and might be changed or removed in future.
  * }</pre>
  */
 @BetaApi("This feature is currently experimental and can change in the future")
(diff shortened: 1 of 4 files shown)


igorbernstein2:
@tonytanger Can you take a look as well? I want to make sure this doesn't conflict with your channel refresh work

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClientFactory.java
```diff
@@ -0,0 +1,202 @@
...
+ *
+ * BigtableDataClientFactory clientFactory = BigtableDataClientFactory.create(defaultSettings);
+ *
+ * // Create a new client for "my-default-instance" in "my-default-project";
+ * BigtableDataClient defaultInstanceClient = clientFactory.createDefault();
+ *
+ * // Create a new client for a different application profile
+ * BigtableDataClient otherAppProfileClient = clientFactory.createForAppProfile("other-app-profile");
+ *
+ * // Create a new client for a completely different instance and application profile.
```
kolea2: what happens if a user puts in an already used project/instance/profile?

igorbernstein2: then they'll get the same result as calling createDefault(). Shouldn't be a problem

kolea2: sgtm, minor nit to add a test for that

igorbernstein2: there is nothing special to test for here...it doesnt matter if the project/instance/profile match the default or not

tonytanger:
LGTM. Don't think primechannel and this will conflict.

PR #115, opened 2019-12-12
feat: add experimental channel refreshing

tonytanger:
Implement ChannelPrimer to establish connection to the GFE so bigtable client will reconnect to the Cloud Bigtable service periodically before the connection gets terminated.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
@@ -184,6 +184,12 @@ public final class BigtableDataSettings {
     return stubSettings.getAppProfileId();
   }
 
+  /** Gets if channels will gracefully refresh connections to Cloud Bigtable service */
+  @BetaApi("This API depends on experimental gRPC APIs")
+  public boolean isRefreshingChannel() {
+    return stubSettings.isRefreshingChannel();
+  }
+
   /** Returns the underlying RPC settings. */
   public EnhancedBigtableStubSettings getStubSettings() {
     return stubSettings;
@@ -273,6 +279,26 @@ public final class BigtableDataSettings {
     /** Gets the CredentialsProvider to use for getting the credentials to make calls with. */
     public CredentialsProvider getCredentialsProvider() {
       return stubSettings.getCredentialsProvider();
+    }
+
+    /**
+     * Configure periodic gRPC channel refreshes.
+     *
+     * <p>This feature will gracefully refresh connections to the Cloud Bigtable service. This is an
+     * experimental feature to address tail latency caused by the service dropping long lived gRPC
+     * connections, which causes the client to renegotiate the gRPC connection in the request path,
+     * which causes periodic spikes in latency
+     */
+    @BetaApi("This API depends on experimental gRPC APIs")
+    public Builder setRefreshingChannel(boolean isRefreshingChannel) {
+      stubSettings.setRefreshingChannel(isRefreshingChannel);
+      return this;
+    }
+
+    /** Gets if channels will gracefully refresh connections to Cloud Bigtable service */
+    @BetaApi("This API depends on experimental gRPC APIs")
+    public boolean isRefreshingChannel() {
+      return stubSettings.isRefreshingChannel();
     }
 
     /**
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RefreshChannel.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RefreshChannel.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.data.v2.internal;
+
+import com.google.api.core.BetaApi;
+import com.google.api.core.InternalApi;
+import com.google.api.gax.grpc.ChannelPrimer;
+import io.grpc.ConnectivityState;
+import io.grpc.ManagedChannel;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Establish a connection to the Cloud Bigtable service on managedChannel
+ *
+ * <p>This class is considered an internal implementation detail and not meant to be used by
+ * applications.
+ */
+@BetaApi("This API depends on gRPC experimental API")
+@InternalApi
+public final class RefreshChannel implements ChannelPrimer {
+
+  /**
+   * primeChannel establishes a connection to Cloud Bigtable service. This typically take less than
+   * 1s. In case of service failure, an upper limit of 10s prevents primeChannel from looping
+   * forever.
+   */
+  @Override
+  public void primeChannel(ManagedChannel managedChannel) {
+    for (int i = 0; i < 10; i++) {
+      ConnectivityState connectivityState = managedChannel.getState(true);
+      if (connectivityState == ConnectivityState.READY) {
+        break;
+      }
+      try {
+        TimeUnit.SECONDS.sleep(1);
+      } catch (InterruptedException e) {
+        break;
+      }
+    }
+  }
+}
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
@@ -15,6 +15,7 @@
  */
 package com.google.cloud.bigtable.data.v2.stub;
 
+import com.google.api.core.BetaApi;
 import com.google.api.gax.batching.BatchingSettings;
 import com.google.api.gax.batching.FlowControlSettings;
 import com.google.api.gax.batching.FlowController.LimitExceededBehavior;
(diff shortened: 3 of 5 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -15,12 +15,15 @@
  */
 package com.google.cloud.bigtable.data.v2;
 
+import com.google.api.client.util.Preconditions;
```
kolea2: Please use `import com.google.common.base.Preconditions;` instead (that should fix the dependency build failure).

igorbernstein2:
As discussed offline
1. the implementation should be moved to EnhancedStubSettings
2. the ChannelProvider meddling should be deferred to the construction of EnhancedBigtableStubSettings. Otherwise the customer has to worry about the ordering of setting the channel provider and enabling channel priming

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -275,6 +275,15 @@ public CredentialsProvider getCredentialsProvider() {
       return stubSettings.getCredentialsProvider();
     }
 
+    /**
+     * Sets to true to enable channels to periodically reconnect to GFE, false (default) to disable
+     */
```
igorbernstein2: Customers don't really know what a GFE is, so I would say "Cloud Bigtable servers". Also I would expand the docs to something like:

```
Configure periodic gRPC channel refreshes.

<p>This feature will gracefully refresh connections to the Cloud Bigtable service. This is an experimental feature to address tail latency caused by the service dropping long lived gRPC connections, which causes the client to renegotiate the gRPC connection in the request path, which causes periodic spikes in latency 

```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RefreshChannel.java
```diff
@@ -0,0 +1,48 @@
...
+import io.grpc.ManagedChannel;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Establish a connection to the GFE on managedChannel
+ *
+ * <p>This class is considered an internal implementation detail and not meant to be used by
+ * applications.
+ */
+@ExperimentalApi("This API depends on gRPC experimental API")
```
igorbernstein2: Please use google api annotations instead:
`com.google.api.core.BetaApi`

Also please make this class final

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RefreshChannel.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.google.api.core.InternalApi;
+import com.google.api.gax.grpc.ChannelPrimer;
+import io.grpc.ConnectivityState;
+import io.grpc.ExperimentalApi;
+import io.grpc.ManagedChannel;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Establish a connection to the GFE on managedChannel
```
igorbernstein2: Please don't reference GFE, customers dont know what that is. Just Cloud Bigtable servers should be good

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -433,6 +443,7 @@ public Builder toBuilder() {
      */
     private Builder() {
       this.appProfileId = SERVER_DEFAULT_APP_PROFILE_ID;
+      this.isRefreshingChannel = false;
```
igorbernstein2: Please update `Builder(EnhancedBigtableStubSettings settings)` to respect this flag as well

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RefreshChannel.java
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * <p>This class is considered an internal implementation detail and not meant to be used by
+ * applications.
+ */
+@ExperimentalApi("This API depends on gRPC experimental API")
+@InternalApi
+public class RefreshChannel implements ChannelPrimer {
+  @Override
+  public void primeChannel(ManagedChannel managedChannel) {
+    for (int i = 0; i < 10; i++) {
```
igorbernstein2: why 10?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -210,6 +214,11 @@ public String getAppProfileId() {
     return appProfileId;
   }
 
+  /** Returns if channels will periodically reconnect to GFE */
+  public boolean isRefreshingChannel() {
+    return isRefreshingChannel;
+  }
```
igorbernstein2: I would drop the getter...I think it will add needless complexity once you fix the toBuilder() bug

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -642,6 +668,16 @@ public EnhancedBigtableStubSettings build() {
       Preconditions.checkState(projectId != null, "Project id must be set");
       Preconditions.checkState(instanceId != null, "Instance id must be set");
 
+      // Set ChannelPrimer on TransportChannelProvider so channels will periodically reconnect to
+      // GFE
```
igorbernstein2: please remove all instances of `GFE`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
igorbernstein2: actually nvm, you already did

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
igorbernstein2: nvm, it already correct

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RefreshChannel.java
tonytanger: It needs an upper limit so it doesn't loop forever. I think 10 seconds should be more than enough time to establish the connection.

igorbernstein2: Please add that as a comment

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -602,6 +614,22 @@ public String getAppProfileId() {
...
+     * @see com.google.cloud.bigtable.data.v2.BigtableDataSettings.Builder#setRefreshingChannel
+     */
+    @BetaApi("This API depends on experimental gRPC APIs")
+    public Builder setRefreshingChannel(boolean isRefreshingChannel) {
+      this.isRefreshingChannel = isRefreshingChannel;
+      return this;
+    }
+
+    /** Gets if channels will gracefully refresh connections to Cloud Bigtable service */
+    public boolean isRefreshingChannel() {
```
igorbernstein2: `@BetaApi`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -275,6 +275,25 @@ public CredentialsProvider getCredentialsProvider() {
...
+     * which causes periodic spikes in latency
+     */
+    @BetaApi("This API depends on experimental gRPC APIs")
+    public Builder setRefreshingChannel(boolean isRefreshingChannel) {
+      stubSettings.setRefreshingChannel(isRefreshingChannel);
+      return this;
+    }
+
+    /** Gets if channels will gracefully refresh connections to Cloud Bigtable service */
+    public boolean isRefreshingChannel() {
```
igorbernstein2: `@BetaApi`

Also please add the getter to BigtableDataSettings as well

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -642,6 +670,16 @@ public EnhancedBigtableStubSettings build() {
       Preconditions.checkState(projectId != null, "Project id must be set");
       Preconditions.checkState(instanceId != null, "Instance id must be set");
 
+      // Set ChannelPrimer on TransportChannelProvider so channels will gracefully refresh
+      // connections to Cloud Bigtable service
+      if (isRefreshingChannel
+          && getTransportChannelProvider() instanceof InstantiatingGrpcChannelProvider) {
```
igorbernstein2: I think this should be a Precondition:

```
if (isRefreshingChannel) {
   Preconditions.checkState(getTransportChannelProvider() instanceof InstantiatingGrpcChannelProvider, "channelRefreshing only works with InstantiatingGrpcChannelProviders")
  ....
}
```

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettingsTest.java
```diff
@@ -61,6 +61,7 @@ public void settingsAreNotLostTest() {
     String projectId = "my-project";
     String instanceId = "my-instance";
     String appProfileId = "my-app-profile-id";
+    boolean isRefreshingChannel = true;
```
kolea2: can you please add a small test for the default (unset) refreshingChannel as well and that it is false?

tonytanger: added 2 tests for default and false

PR #130, opened 2019-12-27
feat: introduce google-cloud-bigtable-deps-bom

igorbernstein2:
Move all of the dependency version management out of the root pom and into a separate BOM.
This will allow java-bigtable-hbase to sync on dependency versions with java-bigtable.

In addition to moving the dependencyManagement section to a BOM, I imported `google-auth-library-bom` and fixed the README to point to the correct version of gRPC

### DIFF
--- a/google-cloud-bigtable-deps-bom/README.md
+++ b/google-cloud-bigtable-deps-bom/README.md
@@ -32,8 +32,8 @@ Example usage:
 </dependencyManagement>
 
 <dependencies>
-  <!-- Declare dependency on google-cloud-client, combined with the above this will ensure a
-    consistent set of dependencies --> 
+  <!-- Declare dependency on google-cloud-client, combined with the above this will
+    ensure a consistent set of dependencies --> 
   <dependency>
     <groupId>com.google.cloud</groupId>
     <artifactId>google-cloud-bigtable</artifactId>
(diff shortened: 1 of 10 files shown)


google-cloud-bigtable-deps-bom/README.md
```diff
@@ -0,0 +1,44 @@
...
+    <!-- Lock google-cloud-bigtable direct dependency versions -->
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigtable-bom</artifactId>
+      <version>1.8.0</version>
+      <type>pom</type>
+      <scope>import</scope>
+    </dependency>
+    <!-- Lock google-cloud-bigtable transitive dependency versions -->
+    <dependency>
```
kolea2: Let me know your thoughts - what if we combined these two into one (similar to how the libraries-bom contains different BOMs), or do you think that's too much?

igorbernstein2: I want both google-cloud-bigtable and java-bigtable-hbase to use the same bom. So it can’t contain a reference for google-cloud-bigtable, otherwise google-cloud-bigtable have itself as a dep

kolea2: Gotcha, thanks for the explanation!

kolea2:
CC @chingor13 to see if there are any last thoughts before I approve here

PR #165, opened 2020-01-21
chore: java doc update to for returned data instances

### LINKED ISSUE #114: Update java docs to clarify thread safety of the client
The docs should make it clear that the client instances can be safely shared across threads. The methods that return non-threadsafe instances should be prominently documented as such on both the creator method and the class returned.

For example, BigtableDataClient#newBulkMutationBatcher() can return a Batcher instance that cannot be shared across threads. The docs for newBulkMutationBatcher() should explicitly state that the returned Batcher instance is not threadsafe, so cannot be used by more than one thread. And in addition, the Batcher interface should also explicitly state in its javadocs that the interface can only be used by a single thread.

rahulKQL:
Fixes #114

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -64,6 +64,9 @@ import javax.annotation.Nullable;
  * an application. However, close() needs to be called on the client object to clean up resources
  * such as threads during application shutdown.
  *
+ * <p>This client can be safely shared across multiple threads except for the Batcher instances
+ * returned from bulk operations, eg. `newBulkMutationBatcher()`, `newBulkReadRowsBatcher()`.
+ *
  * <p>The surface of this class includes several types of Java methods for each of the API's
  * methods:
  *
@@ -911,9 +914,8 @@ public class BigtableDataClient implements AutoCloseable {
 
   /**
    * Mutates multiple rows in a batch. Each individual row is mutated atomically as in MutateRow,
-   * but the entire batch is not executed atomically.
-   *
-   * <p>The returned Batcher instance is not threadsafe, it can only be used from single thread.
+   * but the entire batch is not executed atomically. The returned Batcher instance is not
+   * threadsafe, it can only be used from single thread.
    *
    * <p>Sample Code:
    *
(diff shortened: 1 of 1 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -941,6 +943,8 @@ public void bulkMutateRows(BulkMutation mutation) {
    * Reads rows for given tableId in a batch. If the row does not exist, the value will be null.
    * This operation should be called with in a single thread.
    *
+   * <p>The returned Batcher instance is not threadsafe, it can only be used from single thread.
```
kolea2: I think you can combine this with the previous sentence.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -975,6 +979,8 @@ public void bulkMutateRows(BulkMutation mutation) {
    * Reads rows for given tableId and filter criteria in a batch. If the row does not exist, the
    * value will be null. This operation should be called with in a single thread.
    *
+   * <p>The returned Batcher instance is not threadsafe, it can only be used from single thread.
```
kolea2: same here

kolea2:
also - conventional commits check is failing I believe because there is not a space between `chore:java...`

rahulKQL:
Thanks for the review, I have addressed previous comments, Please have a fresh look.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -64,6 +64,9 @@
  * an application. However, close() needs to be called on the client object to clean up resources
  * such as threads during application shutdown.
  *
+ * <p>This client can be safely shared across multiple threads except for the Batcher instances
+ * returned from bulk operations, eg. `newBulkMutationBatcher()`, `newBulkReadRowsBatcher()`.
+ *
```
igorbernstein2: Also, the returned `ServerStream` from  `readRows` can't be used across multiple threads

rahulKQL: Thanks for the review!!!

I have added similar comment on `BigtableDataClient#readRows()`. PTAL

PR #166, opened 2020-01-23
chore: add filter size check on read row request

rahulKQL:
Currently, bigtable could only serve read row requests with the filter size of 20MB. Adding this check to prevent failure.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
@@ -167,9 +167,11 @@ public final class Query implements Serializable {
    */
   public Query filter(Filters.Filter filter) {
     Preconditions.checkNotNull(filter, "filter can't be null");
+
     RowFilter rowFilter = filter.toProto();
     Preconditions.checkArgument(
         rowFilter.getSerializedSize() < MAX_FILTER_SIZE, "filter size can't be more than 20MB");
+
     builder.setFilter(rowFilter);
     return this;
   }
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
```diff
@@ -162,7 +166,11 @@ public Query range(ByteStringRange range) {
    * filters, please use {@link Filters#interleave()} or {@link Filters#chain()}.
    */
   public Query filter(Filters.Filter filter) {
-    builder.setFilter(filter.toProto());
+    Preconditions.checkNotNull(filter, "filter can't be null");
```
kolea2: can we add a test please?

rahulKQL: Thanks for the review. Added test case for null check. I haven't added any test case for the filter size. I hope that should be fine.

kolea2: Apologies for being unclear, if you could add a test case for filter size as well that would be awesome :)

rahulKQL: I intentionally didn't add test case for that. Reason being the filter size needs to be more than 20MB, I thought that'd be a little large.

I would add unit test right away.

kolea2: Understood - LGTM. @igorbernstein2 any other thoughts?

PR #182, opened 2020-01-31
deps: dependency order matters

elharo:
@kolea2 not going to reorder these right now, but in maven dependency order is significant, and we've been taking advantage of that to fix some dependency issues. We should not promise to sort them alphabetically.

### DIFF
--- a/google-cloud-bigtable/pom.xml
+++ b/google-cloud-bigtable/pom.xml
@@ -51,8 +51,7 @@
   </dependencyManagement>
 
   <dependencies>
-    <!-- NOTE: Dependencies are organized into two groups, production and test.
-         Within a group, dependencies are sorted by (groupId, artifactId) -->
+    <!-- NOTE: Dependencies are organized into two groups, production and test. -->
 
     <!-- Production dependencies -->
     <dependency>


igorbernstein2:
While true that dependency ordering matters in maven,  we don't currently need to rely on dependency ordering. All relevant dependencies are specified in
https://github.com/googleapis/java-bigtable/blob/master/google-cloud-bigtable-deps-bom/pom.xml

If we run into a situation where we start depending on the order, then we can remove the current organization scheme. Alternatively, if you have a preference for a different scheme organization scheme, please let us know and we can migrate to it. But for now, I'd like to keep the long list of prod dependencies organized somehow

elharo:
sorry, this is not that simple. This really does matter.

igorbernstein2:
Please explain

elharo:
suppose two of your dependencies, say http-client and google-auth-library pull in different versions of Apache commons logging at the same depth. Then which one your project gets depends on whether you depend on http-client first and google-auth-library second or vice versa.

This came up in another project recently where we had to reorder dependencies to avoid pulling in an older commons-logging:

https://github.com/googleapis/google-http-java-client/pull/981

Dependency order matters to Maven. Nine times out of ten any arbitrary order you pick will work, but the tenth time you're going to need to pay closer attention than that.

igorbernstein2:
I understand that maven cares about ordering to resolve transitive dependency conflicts and that there is a difference in which version of commons-logging gets pulled in depending on the order of http-client and google-auth-library.
However is this not a problem in this project and to make sure that its not a problem in the future, I'm adding an enforcer rule: https://github.com/googleapis/java-bigtable/pull/203 that will prevent issues like googleapis/google-http-java-client#981.

This project has a lot of dependencies and we need an consistent way to keep them organized. Currently, alphabetical ordering works just fine. If it becomes an issue, we can either add a documented exception to the alphabetical organization or we can change it to another system. I'm not ok from completely removing guidance for organizing dependencies.

PR #203, opened 2020-02-24
chore: ensure that transitive deps have the latest version selected

igorbernstein2:
This should mitigate diamond dependency issues.
Gradle and maven have different behaviors when resolving transitive dependency version conflicts. Gradle selects the greatest version, while maven picks the closest in the dependency hierarchy. For example:
Say http-client and google-auth-library pull in different versions of Apache commons logging at the same depth. Then which one your project gets depends on whether you depend on http-client first and google-auth-library second or vice versa.

This PR enforces that the resolution behavior remains consistent and will break the build.

Fixes issues like googleapis/google-http-java-client#981

### DIFF
--- a/google-cloud-bigtable/pom.xml
+++ b/google-cloud-bigtable/pom.xml
@@ -366,7 +366,7 @@
           </execution>
           <!-- Ensure that greatest version of each transitive dependency got selected -->
           <execution>
-            <id>enforce-dependencu-upper-bound</id>
+            <id>enforce-dependency-upper-bound</id>
             <configuration>
               <rules>
                 <requireUpperBoundDeps/>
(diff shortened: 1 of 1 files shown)


google-cloud-bigtable/pom.xml
```diff
@@ -363,6 +364,18 @@
               <goal>enforce</goal>
             </goals>
           </execution>
+          <!-- Ensure that greatest version of each transitive dependency got selected -->
+          <execution>
+            <id>enforce-dependencu-upper-bound</id>
```
elharo: dependency

igorbernstein2: thanks, fixed

PR #221, opened 2020-03-18
chore: clean up synth.py script with newer tooling

kolea2:


synth.py
```diff
@@ -86,32 +69,16 @@ def generate_data_api(gapic, license):
...
-  java.fix_proto_headers(library / f'proto-google-cloud-bigtable-v2')
-  # java.fix_grpc_headers(library / f'grpc-google-cloud-bigtable-v2', package_name)
-
-  ##todo replace with fix_grpc_headers
-  s.replace(
-      library / f'grpc-google-cloud-bigtable-v2/src/**/*.java',
-      f"package com.google.bigtable.v2;",
-      f"{license}package com.google.bigtable.v2;"
-  )
-  s.copy(library / f'gapic-google-cloud-bigtable-v2', 'google-cloud-bigtable/', excludes=excludes)
```
chingor13: The `java.gapic_library()` helper won't exclude these files

synth.py
```diff
@@ -86,32 +69,16 @@ def generate_data_api(gapic, license):
...
-      config_path='/google/bigtable/admin/artman_bigtableadmin.yaml',
-      artman_output_name='')
+def generate_admin_api(gapic):
+  java.gapic_library(
+        service='bigtable-admin',
+        version='v2',
+        config_pattern='/google/bigtable/admin/artman_bigtableadmin.yaml',
+        package_pattern='com.google.bigtable.v2',
+        gapic=gapic,
+  )
```
igorbernstein2: I don't think this will put these files in the right place. java-bigtable is a single artifact that contains clients for both admin & data. Which is why all of the explicit copies were added

PR #232, opened 2020-03-26
chore: Migrate bigtable synth.py to bazel

vam-google:


vam-google:
@chingor13 PTAL

synth.py
```diff
@@ -70,12 +73,12 @@ def generate_data_api(gapic):
   ]
 
   package_name = f'com.google.bigtable.v2'
-  java.fix_proto_headers(library / f'proto-google-cloud-bigtable-v2')
-  java.fix_grpc_headers(library / f'grpc-google-cloud-bigtable-v2', package_name)
+  java.fix_proto_headers(library / f'proto-google-cloud-bigtable-v2-java')
```
kolea2: I don't think these should be changing. This should have generated at least the changes found here: https://github.com/googleapis/java-bigtable/pull/222 which we have yet to merge in.

vam-google: Sorry, that was silly me. I updated the synth.py, but did not update my command to call synthtool, so I just generated java-gerafeas (the repo I was migrating before java-bigtable).

vam-google:
@kolea2 PTAL. I regenerated the files, and now it seems like there are a lot of changes, including breaking ones =(.

chingor13:
As an FYI, we added a `required` option to `s.copy()`/`s.move()` which will cause the command to fail if nothing was copied (as a sanity check)

kolea2:
@vam-google thanks! It looks like the same tests are failing in the other PR, I think we need to update a test. I'll take a look and let you know when this is good to come back to. Sorry for the inconvenience!

vam-google:
@kolea2 Friendly ping

vam-google:
Thanks @kolea2 Can we now merge it? (I don't have right to merge)

PR #249, opened 2020-04-15
chore: add IT args to nightly builds

kolea2:


.kokoro/nightly/integration.cfg
```diff
@@ -6,6 +6,11 @@ env_vars: {
   value: "gcr.io/cloud-devrel-kokoro-resources/java8"
 }
 
+env_vars: {
+    key: "INTEGRATION_TEST_ARGS"
+    value: "-P bigtable-emulator-it,bigtable-prod-it -Dbigtable.project=gcloud-devel -Dbigtable.instance=google-cloud-bigtable -Dbigtable.table=integration-tests"
```
gguuss, line 11: Should these be stored in the environment variables?

kolea2: This is how we invoke IT tests for presubmits - we can take a look at setting these elsewhere, but for now I'd like them to be consistent. https://github.com/googleapis/java-bigtable/blob/master/.kokoro/presubmit/integration.cfg#L9-L12

gguuss: Ack. LGTM then :)

PR #281, opened 2020-05-05
feat: add more context to row merging errors

igorbernstein2:
These errors should never happen. But if they do, it would very useful to get more context as to what rows triggered them


igorbernstein2:
Should be ready now

PR #284, opened 2020-05-06
chore: Create CODEOWNERS

czahedi:
Adding owner team

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,4 +7,4 @@
 
 # The bigtable-dpe team is the default owner for anything not
 # explicitly taken by someone else.
-*                               @GoogleCloudPlatform/bigtable-dpe
+*                               @googleapis/bigtable-dpe
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,10 @@
+# Code owners file.
+# This file controls who is tagged for review for any given pull request.
+#
+# For syntax help see:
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1,10 @@
+# Code owners file.
+# This file controls who is tagged for review for any given pull request.
+#
+# For syntax help see:
+# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+
+
+# The bigtable-dpe team is the default owner for anything not
+# explicitly taken by someone else.
+*                               @GoogleCloudPlatform/bigtable-dpe
```
igorbernstein2, line 10: please add me as well

chingor13, line 10: As far as I know, we can't grant access to teams in a different GitHub organization.

czahedi: Is that because of auth issues or because of our policy?

I can make a googleapis@ version of the teams either way, but if it's for policy I worry about creating two different teams and needed to check them for access / removing access.

czahedi: I need to make googleapis@ versions of the teams per Jeff's comment below. Will amend this and just add you to both teams.

czahedi: Added you to both teams.

czahedi: Made the correct team and adjusted the owners file.

PR #293, opened 2020-05-08
fix: set x-goog-api-client as internal header

kolea2:


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/HeadersTest.java
```diff
@@ -104,6 +105,11 @@ public void setUp() throws Exception {
                 .setElementCountThreshold(1L)
                 .build());
 
+    // getInternalHeaderProvider is a protected method so we can't call it, instead ensure that
+    // headers were not set here. the test will assert the full header later on
+    if (!(settings.stubSettings().getHeaderProvider() instanceof NoHeaderProvider)) {
+      throw new AssertionError("header provider should not be set");
+    }
```
igorbernstein2, line 112: I dont love this:
- we shouldnt test things in setup()...it should only setup the environment for the test
- I dont think enforcing that header provider is unset is appropriate.

I think it would be better to set a static header provider here and test that it doesnt clobber the other headers instead

kolea2:
@igorbernstein2 updated

PR #299, opened 2020-05-11
feat: adding utility to transform protobuf into model object

### LINKED ISSUE #297: Create fromProtobuf utility method in existing models
**Please describe requested feature? ** 
User should be able to transform protobuf object into bigtable client's model object. Similar to:
https://github.com/googleapis/java-bigtable/blob/43744924be03ee862d071be6cc6d07683945aa9d/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java#L272-L285

**Additional context**
Discussed offline in the last meeting.

rahulKQL:
Fixes #297

With this commit, the User shall be able to transform protobuf object to bigtable client's specific model objects.

- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-bigtable/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
@@ -118,11 +118,16 @@ public final class BulkMutation implements Serializable, Cloneable {
   /**
    * Wraps the protobuf {@link MutateRowsRequest}.
    *
-   * <p>WARNING: Please note that the table name will be overwritten by the configuration in the
-   * BigtableDataClient. The mutation must always be idempotent because it would be retried.
+   * <p>This is meant for advanced usage only. Please ensure that the MutateRowsRequest does not use
+   * server side timestamps. The BigtableDataClient assumes that mutation present in BulkMutation
+   * are idempotent and is configured to enable retries by default. If serverside timestamps are
+   * enabled then that can lead to duplicate mutations.
+   *
+   * <p>WARNING: when applied, the resulting mutation object will ignore the project id and instance
+   * id in the table_name and instead apply the configuration in the client.
    */
   @BetaApi
-  public static BulkMutation fromProtobuf(@Nonnull MutateRowsRequest request) {
+  public static BulkMutation fromProto(@Nonnull MutateRowsRequest request) {
     BulkMutation bulkMutation =
         BulkMutation.create(NameUtil.extractTableIdFromTableName(request.getTableName()));
     bulkMutation.builder = request.toBuilder();
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
@@ -114,6 +114,20 @@ public final class BulkMutation implements Serializable, Cloneable {
         .build();
   }
 
+  /**
+   * Wraps the protobuf {@link MutateRowsRequest}.
+   *
+   * <p>WARNING: Please note that the project id & instance id in the table name will be overwritten
+   * by the configuration in the BigtableDataClient.
+   */
+  public static BulkMutation fromProtobuf(@Nonnull MutateRowsRequest request) {
+    BulkMutation bulkMutation =
+        BulkMutation.create(NameUtil.extractTableIdFromTableName(request.getTableName()));
+    bulkMutation.builder = request.toBuilder();
+
+    return bulkMutation;
+  }
+
   /** Creates a copy of {@link BulkMutation}. */
   @Override
   public BulkMutation clone() {
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
@@ -15,6 +15,7 @@
  */
 package com.google.cloud.bigtable.data.v2.models;
 
+import com.google.api.core.BetaApi;
 import com.google.api.core.InternalApi;
 import com.google.bigtable.v2.MutateRowsRequest;
 import com.google.cloud.bigtable.data.v2.internal.NameUtil;
@@ -117,9 +118,10 @@ public final class BulkMutation implements Serializable, Cloneable {
   /**
    * Wraps the protobuf {@link MutateRowsRequest}.
    *
-   * <p>WARNING: Please note that the project id & instance id in the table name will be overwritten
-   * by the configuration in the BigtableDataClient.
+   * <p>WARNING: Please note that the table name will be overwritten by the configuration in the
+   * BigtableDataClient. The mutation must always be idempotent because it would be retried.
    */
+  @BetaApi
   public static BulkMutation fromProtobuf(@Nonnull MutateRowsRequest request) {
     BulkMutation bulkMutation =
         BulkMutation.create(NameUtil.extractTableIdFromTableName(request.getTableName()));
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ConditionalRowMutation.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ConditionalRowMutation.java
@@ -15,6 +15,7 @@
  */
 package com.google.cloud.bigtable.data.v2.models;
 
+import com.google.api.core.BetaApi;
 import com.google.api.core.InternalApi;
 import com.google.bigtable.v2.CheckAndMutateRowRequest;
 import com.google.cloud.bigtable.data.v2.internal.NameUtil;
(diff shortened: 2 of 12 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
```diff
@@ -114,6 +114,20 @@ public MutateRowsRequest toProto(RequestContext requestContext) {
         .build();
   }
 
+  /**
+   * Wraps the protobuf {@link MutateRowsRequest}.
+   *
+   * <p>WARNING: Please note that the project id & instance id in the table name will be overwritten
```
igorbernstein2, line 120: just the table name

igorbernstein2: Also please add a warning that the mutations should be idempotent (ie should not use server timestamps) because the mutations will be retried

igorbernstein2: Also, lets mark this as BetaApi, in case this doesn't prove useful

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ConditionalRowMutation.java
```diff
@@ -136,4 +136,19 @@ public CheckAndMutateRowRequest toProto(RequestContext requestContext) {
         .setAppProfileId(requestContext.getAppProfileId())
         .build();
   }
+
+  /**
+   * Wraps the protobuf {@link CheckAndMutateRowRequest}.
+   *
+   * <p>WARNING: Please note that the project id & instance id in the table name will be overwritten
+   * by the configuration in the BigtableDataClient.
+   */
```
igorbernstein2, line 145: Same notes as BulkMutation, but you dont need to comment about idempotency this rpc is not retried

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ReadModifyWriteRow.java
```diff
@@ -137,4 +137,19 @@ public ReadModifyWriteRowRequest toProto(RequestContext requestContext) {
         .setAppProfileId(requestContext.getAppProfileId())
         .build();
   }
+
+  /**
+   * Wraps the protobuf {@link ReadModifyWriteRowRequest}.
+   *
+   * <p>WARNING: Please note that the project id & instance id in the table name will be overwritten
+   * by the configuration in the BigtableDataClient.
+   */
```
igorbernstein2, line 146: same as ConditionalRowMutation

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutation.java
```diff
@@ -214,4 +214,18 @@ public MutateRowsRequest toBulkProto(RequestContext requestContext) {
             Entry.newBuilder().setRowKey(key).addAllMutations(mutation.getMutations()).build())
         .build();
   }
+
+  /**
+   * Wraps the protobuf {@link MutateRowRequest}.
+   *
+   * <p>WARNING: Please note that the project id & instance id in the table name will be overwritten
+   * by the configuration in the BigtableDataClient.
```
igorbernstein2, line 222: Same as bulk mutation

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutationEntry.java
```diff
@@ -177,4 +178,11 @@ public RowMutationEntry deleteRow() {
         .addAllMutations(mutation.getMutations())
         .build();
   }
+
+  /** Wraps the protobuf {@link MutateRowRequest}. */
+  public static RowMutationEntry fromProtobuf(MutateRowsRequest.Entry request) {
+    // TODO: Should we have an option for Mutation.fromProto(List<com.google.bigtable.v2.Mutation>)?
+    return new RowMutationEntry(
+        request.getRowKey(), Mutation.fromProtoUnsafe(request.getMutationsList()));
+  }
```
igorbernstein2, line 187: I dont think we need this...its only used by BulkMutation and if someone is using BulkMutation, they are already using the models

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
rahulKQL: Thanks for the review, I have updated the JavaDoc with the same information. PTAL

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ConditionalRowMutation.java
```diff
@@ -136,4 +137,20 @@ public CheckAndMutateRowRequest toProto(RequestContext requestContext) {
...
   }
+
+  /**
+   * Wraps the protobuf {@link CheckAndMutateRowRequest}.
+   *
+   * <p>WARNING: Please note that the table name will be overwritten by the configuration in the
+   * BigtableDataClient. The mutation must always be idempotent because it would be retried.
+   */
+  @BetaApi
+  public static ConditionalRowMutation fromProtobuf(@Nonnull CheckAndMutateRowRequest request) {
```
igorbernstein2, line 148: Here and everywhere else, please stay consistent with existing `fromProto` methods

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -87,6 +87,17 @@ public static Mutation fromProtoUnsafe(List<com.google.bigtable.v2.Mutation> pro
     return mutation;
   }
 
+  /**
+   * Wraps the List of protobuf {@link com.google.bigtable.v2.Mutation}. This method creates a
+   * mutation instance that is idempotent in nature, which could safely be retried.
```
igorbernstein2, line 92: This javadoc is not quite right. Please rephrase to:

```
Constructs a row mutation from an existing protobuf object.

<p>Callers must ensure that the protobuf argument is not using serverside timestamps. The client assumes that all mutations are idempotent and will retry in case of transient errors. This can lead to row duplication.

<p>When applied, the resulting Mutation object will ignore the project id and instance id in the table_name and instead apply the configuration in the client
```

igorbernstein2: Actually nevermind, please remove this fromProto method altogether. This object is not a top level model so from proto is not needed here

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutation.java
```diff
@@ -214,4 +215,18 @@ public MutateRowsRequest toBulkProto(RequestContext requestContext) {
             Entry.newBuilder().setRowKey(key).addAllMutations(mutation.getMutations()).build())
         .build();
   }
+
+  /**
+   * Wraps the protobuf {@link MutateRowRequest}.
+   *
+   * <p>WARNING: Please note that the table name will be overwritten by the configuration in the
+   * BigtableDataClient. The mutation must always be idempotent because it would be retried.
```
igorbernstein2, line 223: Please replace with:

```
<p>Please ensure that the MutateRowRequest does not use server side timestamps. The BigtableDataClient assumes that RowMutations are idempotent and is configured to enable retries by default. If serverside timestamps are enabled, this can lead to duplicate mutations
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
```diff
@@ -114,6 +115,21 @@ public MutateRowsRequest toProto(RequestContext requestContext) {
         .build();
   }
 
+  /**
+   * Wraps the protobuf {@link MutateRowsRequest}.
+   *
+   * <p>WARNING: Please note that the table name will be overwritten by the configuration in the
+   * BigtableDataClient. The mutation must always be idempotent because it would be retried.
```
igorbernstein2, line 122: `The mutation must always be idempotent because it would be retried.` needs to be explained more, please add a similar warning as described below for RowMutation.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
rahulKQL: It seems we need this utility for `RowMutation#fromProto`. At present, there is no other option to create `model.Mutation` from List of proto `com.google.bigtable.v2.Mutation`.

igorbernstein2: ok, can it be package private? or internalapi?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Mutation.java
```diff
@@ -87,6 +88,23 @@ public static Mutation fromProtoUnsafe(List<com.google.bigtable.v2.Mutation> pro
...
+   * Constructs a row mutation from an existing protobuf object.
+   *
+   * <p>Callers must ensure that the protobuf argument is not using serverside timestamps. The
+   * client assumes that all mutations are idempotent and will retry in case of transient errors.
+   * This can lead to row duplication.
+   *
+   * <p>When applied, the resulting Mutation object will ignore the project id and instance id in
+   * the table_name and instead apply the configuration in the client
+   */
+  @VisibleForTesting
```
igorbernstein2, line 101: Why is this visible for testing? I thought it was needed by RowMutation?

rahulKQL: This was my overlook, I have removed the unneeded annotation. PTAL

PR #313, opened 2020-05-18
chore(fix): unflake MetricsTracerTest

### LINKED ISSUE #310: Investigate flaky MetricsTracerTest
https://github.com/googleapis/java-bigtable/runs/685522354#step:5:3538

kolea2:
Fixes #310

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/MetricsTracerTest.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/MetricsTracerTest.java
@@ -206,9 +206,8 @@ public class MetricsTracerTest {
 
   @Test
   public void testReadRowsFirstRow() throws InterruptedException {
-    final long beforeSleep = 40;
-    final long afterSleep = 40;
-
+    final long beforeSleep = 60;
+    final long afterSleep = 60;
 
     doAnswer(
             new Answer() {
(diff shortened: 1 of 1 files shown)


igorbernstein2:
I dont think this will fix the flakiness. I think the issue is that there is a race between the server thread (the emulator) and the client recording the duration:

client: starts 2 timers (first response, overall op) and sends request
server: sleeps(beforeSleep), sends response
> race starts now
server: sleeps (afterSleep)
client: stops first response timer

server: finishes the call
client: stops overall timer

if the client won the race then timer will not be greater than  overall timer - afterSleep, however if the server won the race then this condition will not be true.

I think the easiest fix is to add some slop to:
assertThat(firstRowLatency).isIn(Range.closed(beforeSleep, elapsed - afterSleep));

something like:
assertThat(firstRowLatency).isIn(Range.closed(beforeSleep, elapsed - afterSleep/2));

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/MetricsTracerTest.java
```diff
@@ -237,7 +237,7 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
         getAggregationValueAsLong(
             RpcViewConstants.BIGTABLE_READ_ROWS_FIRST_ROW_LATENCY_VIEW,
             ImmutableMap.<TagKey, TagValue>of());
-    assertThat(firstRowLatency).isIn(Range.closed(beforeSleep, elapsed - afterSleep));
+    assertThat(firstRowLatency).isIn(Range.closed(beforeSleep, elapsed - afterSleep / 2));
```
igorbernstein2, line 240: Please add a comment explaining whats happening :)

PR #335, opened 2020-06-09
Revert "fix: change app_profile_id to the correct header app_profile"

igorbernstein2:
Reverts googleapis/java-bigtable#318

The header keys should match the request fields


tonytanger:
LGTM!

PR #350, opened 2020-06-17
feat(deps): adopt flatten plugin and google-cloud-shared-dependencies

yangnuoyu:


igorbernstein2:
Please attach a diff of how the dependencies change for end users

stephaniewang526:
> Please attach a diff of how the dependencies change for end users

@igorbernstein2 there is no change in dependency list for end users -- i added a kokoro check for this in dependencies.sh to ensure accuracy. The only difference is that transitive dependencies are resolved to be direct. i will share my doc with you for further details. Thanks.

igorbernstein2:
Looking at the main artifact, the proposed change will affect the following deps (by pulling them up to the top level). 

```
[INFO] |  \- javax.annotation:javax.annotation-api:jar:1.3.2:compile

[INFO] |  +- com.google.http-client:google-http-client:jar:1.36.0:compile
[INFO] |  |  +- org.apache.httpcomponents:httpclient:jar:4.5.12:compile
[INFO] |  |  |  +- commons-logging:commons-logging:jar:1.2:compile
[INFO] |  |  |  \- commons-codec:commons-codec:jar:1.11:compile
[INFO] |  |  +- org.apache.httpcomponents:httpcore:jar:4.4.13:compile
[INFO] |  |  \- io.opencensus:opencensus-contrib-http-util:jar:0.24.0:compile
[INFO] |  \- com.google.http-client:google-http-client-jackson2:jar:1.36.0:compile
[INFO] |     \- com.fasterxml.jackson.core:jackson-core:jar:2.11.1:compile

[INFO] |  +- com.google.guava:failureaccess:jar:1.0.1:compile
[INFO] |  +- com.google.guava:listenablefuture:jar:9999.0-empty-to-avoid-conflict-with-guava:compile
[INFO] |  +- org.checkerframework:checker-compat-qual:jar:2.5.5:compile
[INFO] |  +- com.google.errorprone:error_prone_annotations:jar:2.4.0:compile
[INFO] |  \- com.google.j2objc:j2objc-annotations:jar:1.3:compile

[INFO] |  \- com.google.code.gson:gson:jar:2.8.6:compile
[INFO] |  +- org.apache.commons:commons-lang3:jar:3.5:runtime
[INFO] |  \- org.conscrypt:conscrypt-openjdk-uber:jar:2.2.1:runtime

[INFO] |  \- org.codehaus.mojo:animal-sniffer-annotations:jar:1.19:runtime
```

This means that those libraries might change version in customer applications if bigtable was tied for a version. But I dont see any tangible negative effects that this would have.

I think we can probably merge

PR #371, opened 2020-07-21
chore: Introspect peer IP address for directpath tests

WeiranFang:
As we are rolling out DirectPath IPv4 support, we want to test both ipv4 and ipv6 cases (i.e. run test client on both ipv4-only VM and ipv6-enabled VM). And to make sure DirectPath is working as expected on both ipv4/ipv6 connection, we would want to introspect the peer IP address after connection is successfully established.

Changes in this PR:
* Added client interceptor in `CloudEnv.java`, which is only enabled for directpath tests. Please advise if there's a better place for adding this interceptor.
* Throws a `RuntimeException` when remote IP address is not as expected.
* Added `bigtable-directpath-ipv4-it` test profile to be run on a ipv4-only VM. The original `bigtable-directpath-it` will still be run on a ipv6-enabled VM.

### DIFF
--- a/google-cloud-bigtable/pom.xml
+++ b/google-cloud-bigtable/pom.xml
@@ -338,6 +338,44 @@
         </plugins>
       </build>
     </profile>
+
+    <profile>
+      <id>bigtable-directpath-ipv4-it</id>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-failsafe-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>directpath-it</id>
+                <goals>
+                  <goal>integration-test</goal>
+                  <goal>verify</goal>
+                </goals>
+                <configuration>
+                  <skip>false</skip>
+
+                  <systemPropertyVariables>
+                    <bigtable.env>cloud</bigtable.env>
+                    <bigtable.data-endpoint>${bigtable.directpath-data-endpoint}</bigtable.data-endpoint>
+                    <bigtable.admin-endpoint>${bigtable.directpath-admin-endpoint}</bigtable.admin-endpoint>
+                    <bigtable.grpc-log-dir>${project.build.directory}/test-grpc-logs/directpath-it</bigtable.grpc-log-dir>
+                    <bigtable.attempt-directpath>true</bigtable.attempt-directpath>
+                    <bigtable.directpath-ipv4>true</bigtable.directpath-ipv4>
+                  </systemPropertyVariables>
+                  <includes>
+                    <!-- TODO(igorbernstein): Once the control plane is accessible via directpath, add admin tests -->
+                    <include>com.google.cloud.bigtable.data.v2.it.*IT</include>
+                  </includes>
+                  <summaryFile>${project.build.directory}/failsafe-reports/failsafe-summary-directpath-it.xml</summaryFile>
+                  <reportsDirectory>${project.build.directory}/failsafe-reports/directpath-ipv4-it</reportsDirectory>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
   </profiles>
 
   <build>
--- a/google-cloud-bigtable/pom.xml
+++ b/google-cloud-bigtable/pom.xml
@@ -359,7 +359,7 @@
                     <bigtable.env>cloud</bigtable.env>
                     <bigtable.data-endpoint>${bigtable.directpath-data-endpoint}</bigtable.data-endpoint>
                     <bigtable.admin-endpoint>${bigtable.directpath-admin-endpoint}</bigtable.admin-endpoint>
-                    <bigtable.grpc-log-dir>${project.build.directory}/test-grpc-logs/directpath-it</bigtable.grpc-log-dir>
+                    <bigtable.grpc-log-dir>${project.build.directory}/test-grpc-logs/directpath-ipv4-it</bigtable.grpc-log-dir>
                     <bigtable.attempt-directpath>true</bigtable.attempt-directpath>
                     <bigtable.directpath-ipv4>true</bigtable.directpath-ipv4>
                   </systemPropertyVariables>
@@ -367,7 +367,7 @@
                     <!-- TODO(igorbernstein): Once the control plane is accessible via directpath, add admin tests -->
                     <include>com.google.cloud.bigtable.data.v2.it.*IT</include>
                   </includes>
-                  <summaryFile>${project.build.directory}/failsafe-reports/failsafe-summary-directpath-it.xml</summaryFile>
+                  <summaryFile>${project.build.directory}/failsafe-reports/failsafe-summary-directpath-ipv4-it.xml</summaryFile>
                   <reportsDirectory>${project.build.directory}/failsafe-reports/directpath-ipv4-it</reportsDirectory>
                 </configuration>
               </execution>
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
@@ -15,6 +15,9 @@
  */
 package com.google.cloud.bigtable.test_helpers.env;
 
+import com.google.api.core.ApiFunction;
+import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
+import com.google.api.gax.rpc.TransportChannelProvider;
 import com.google.cloud.bigtable.admin.v2.BigtableInstanceAdminClient;
 import com.google.cloud.bigtable.admin.v2.BigtableInstanceAdminSettings;
 import com.google.cloud.bigtable.admin.v2.BigtableTableAdminClient;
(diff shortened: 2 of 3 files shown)


WeiranFang:
+cc @igorbernstein2, @apolcyn

kolea2:
Adding do not merge label until 1.14.0 release is complete (starting now)

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
```diff
@@ -81,6 +103,24 @@ private CloudEnv(
       dataSettings.stubSettings().setEndpoint(dataEndpoint);
     }
 
+    if (isDirectPathEnabled()) {
+      TransportChannelProvider channelProvider = dataSettings.stubSettings().getTransportChannelProvider();
+      InstantiatingGrpcChannelProvider defaultTransportProvider = (InstantiatingGrpcChannelProvider) channelProvider;
+      InstantiatingGrpcChannelProvider instrumentedTransportChannelProvider =
+          defaultTransportProvider
+              .toBuilder()
+              .setChannelConfigurator(
```
igorbernstein2, line 112: I think it would be slightly cleaner to use `setInterceptorProvider` instead of the configurator

WeiranFang: Thanks for the pointer. One concern I had of using `setInterceptorProvider` is, if the existing channel provider already has its interceptor provider, then we would be replacing the existing one with the one we created. It would be good if we could retrieve the existing interceptor provider and add our own interceptor on top of that, but I couldn't find a good way of doing this. WDYT?

igorbernstein2: Thats a fair point. It would be nice if gax had the ability to add an interceptor. I guess leave it as is for now

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
```diff
@@ -51,6 +72,7 @@
   private final BigtableDataSettings.Builder dataSettings;
   private final BigtableTableAdminSettings.Builder tableAdminSettings;
   private final BigtableInstanceAdminSettings.Builder instanceAdminSettings;
+  private final AtomicReference<ClientCall<?, ?>> clientCallCapture = new AtomicReference<>();
```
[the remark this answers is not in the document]
WeiranFang: Right, since it's a per-call interceptor, it should be safe to just use a local variable. Thanks for pointing that out

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
```diff
@@ -153,4 +193,53 @@ private static String getRequiredProperty(String prop) {
     }
     return value;
   }
+
+  /**
+   * Captures the request attributes "Grpc.TRANSPORT_ATTR_REMOTE_ADDR" when connection is established.
+   * This is useful for DirectPath testing and debugging.
+   * {@link ClientCall#getAttributes()}
+   */
+  private ClientInterceptor recordClientCallInterceptor(
```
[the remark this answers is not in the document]
WeiranFang: Updated.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
```diff
@@ -153,4 +193,53 @@ private static String getRequiredProperty(String prop) {
...
+        return new SimpleForwardingClientCall<ReqT, RespT>(clientCall) {
+          @Override
+          public void start(Listener<RespT> responseListener, Metadata headers) {
+            super.start(new SimpleForwardingClientCallListener<RespT>(responseListener) {
+              @Override
+              public void onHeaders(Metadata headers) {
+                // Check peer IP after connection is established.
+                SocketAddress remoteAddr = clientCallCapture.get().getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR);
+                if (!verifyRemoteAddress(remoteAddr)) {
+                  throw new RuntimeException(String.format(
```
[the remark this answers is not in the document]
WeiranFang: Yes, it will throw `java.util.concurrent.ExecutionException: com.google.api.gax.rpc.CancelledException: io.grpc.StatusRuntimeException: CANCELLED: Failed to read headers`

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
```diff
@@ -153,4 +193,53 @@ private static String getRequiredProperty(String prop) {
...
+   * {@link ClientCall#getAttributes()}
+   */
+  private ClientInterceptor recordClientCallInterceptor(
+      final AtomicReference<ClientCall<?, ?>> clientCallCapture) {
+    return new ClientInterceptor() {
+      @Override
+      public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+          MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+        ClientCall<ReqT, RespT> clientCall = next.newCall(method,callOptions);
+        clientCallCapture.set(clientCall);
```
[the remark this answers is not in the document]
WeiranFang: Done.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
```diff
@@ -153,4 +191,54 @@ private static String getRequiredProperty(String prop) {
...
+   * {@link ClientCall#getAttributes()}
+   */
+  private ClientInterceptor directPathAddressCheckInterceptor() {
+    return new ClientInterceptor() {
+      private ClientCall<?, ?> capturedCall;
+      @Override
+      public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+          MethodDescriptor<ReqT, RespT> method, CallOptions callOptions, Channel next) {
+        ClientCall<ReqT, RespT> clientCall = next.newCall(method,callOptions);
+        capturedCall = clientCall;
```
igorbernstein2, line 210: I dont think you need capturedCall here, the inner class should be able to access clientCall directly

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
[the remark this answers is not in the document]
igorbernstein2: Good to know that its safe, but the test failure will be hard for someone to track down.
Can you make the message more explicit, something along the lines of:

"Synthetically aborting the current request because it did not adhere to the test environment's requirement for DirectPath. Expected test to access DirectPath via %s, but RPC was destined for %s"

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
WeiranFang: Done.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/CloudEnv.java
```diff
@@ -153,4 +193,57 @@ private static String getRequiredProperty(String prop) {
...
+                    SocketAddress remoteAddr =
+                        clientCall.getAttributes().get(Grpc.TRANSPORT_ATTR_REMOTE_ADDR);
+                    if (!verifyRemoteAddress(remoteAddr)) {
+                      throw new RuntimeException(
+                          String.format(
+                              "Synthetically aborting the current request because it did not adhere"
+                                  + " to the test environment's requirement for DirectPath."
+                                  + " Expected test to access DirectPath via %s,"
+                                  + " but RPC was destined for %s",
+                              remoteAddr.toString(), isDirectPathIpv4() ? "ipv4" : "ipv6"));
```
igorbernstein2, line 225: I think your parameters are backwards:
`Expected test to access DirectPath via %s, but RPC was destined for %s"`
so the values should be flipped:
`isDirectPathIpv4() ? "ipv4" : "ipv6", remoteAddr.toString()`

WeiranFang: oops :) Done.

google-cloud-bigtable/pom.xml
```diff
@@ -338,6 +338,44 @@
...
+                    <bigtable.admin-endpoint>${bigtable.directpath-admin-endpoint}</bigtable.admin-endpoint>
+                    <bigtable.grpc-log-dir>${project.build.directory}/test-grpc-logs/directpath-it</bigtable.grpc-log-dir>
+                    <bigtable.attempt-directpath>true</bigtable.attempt-directpath>
+                    <bigtable.directpath-ipv4>true</bigtable.directpath-ipv4>
+                  </systemPropertyVariables>
+                  <includes>
+                    <!-- TODO(igorbernstein): Once the control plane is accessible via directpath, add admin tests -->
+                    <include>com.google.cloud.bigtable.data.v2.it.*IT</include>
+                  </includes>
+                  <summaryFile>${project.build.directory}/failsafe-reports/failsafe-summary-directpath-it.xml</summaryFile>
```
igorbernstein2, line 370: sorry, just noticed this, I think this has been its own summary file so that it wont clobber the regular directpath-it tests

WeiranFang: Good catch. Also fixed grpc-log-dir.

PR #376, opened 2020-08-03
samples: migrate samples from GoogleCloudPlatform/java-docs-samples /bigtable/snippets

chingor13:


kolea2:
@billyjacobson

.kokoro/nightly/samples.cfg
```diff
@@ -36,3 +36,8 @@ env_vars: {
   key: "ENABLE_BUILD_COP"
   value: "true"
 }
+
+env_vars: {
```
chingor13, line 40: Probably want to add these 2 config files to the template excludes in synth.py

PR #397, opened 2020-08-28
fix: Filters should be serializable

### LINKED ISSUE #385: Filters should be serializable
This would facilitate the ability to write parameterized DoFns for beam

dmitry-fa:
Fixes #385

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Filters.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Filters.java
@@ -200,6 +200,7 @@ public final class Filters {
   // Implementations of target specific filters.
   /** DSL for adding filters to a chain. */
   public static final class ChainFilter implements Filter {
+    private static final long serialVersionUID = -89237431180618430L;
     private RowFilter.Chain.Builder builder;
 
     private ChainFilter() {
@@ -241,6 +242,7 @@ public final class Filters {
 
   /** DSL for adding filters to the interleave list. */
   public static final class InterleaveFilter implements Filter {
+    private static final long serialVersionUID = -3866166430679348474L;
     private RowFilter.Interleave.Builder builder;
 
     private InterleaveFilter() {
@@ -281,6 +283,7 @@ public final class Filters {
 
   /** DSL for configuring a conditional filter. */
   public static final class ConditionFilter implements Filter {
+    private static final long serialVersionUID = -7576541105323990049L;
     private RowFilter.Condition.Builder builder;
 
     private ConditionFilter(@Nonnull Filter predicate) {
@@ -323,7 +326,9 @@ public final class Filters {
     }
   }
 
-  public static final class KeyFilter {
+  public static final class KeyFilter implements Serializable {
+    private static final long serialVersionUID = 5137765114285539458L;
+
     private KeyFilter() {}
 
     /**
@@ -383,7 +388,9 @@ public final class Filters {
     }
   }
 
-  public static final class FamilyFilter {
+  public static final class FamilyFilter implements Serializable {
+    private static final long serialVersionUID = -4470936841191831553L;
+
     private FamilyFilter() {}
 
     /**
@@ -405,7 +412,9 @@ public final class Filters {
     }
   }
 
-  public static final class QualifierFilter {
+  public static final class QualifierFilter implements Serializable {
+    private static final long serialVersionUID = -1274850022909506559L;
+
     private QualifierFilter() {}
 
     /**
@@ -459,7 +468,8 @@ public final class Filters {
 
   /** Matches only cells from columns within the given range. */
   public static final class QualifierRangeFilter
-      extends AbstractByteStringRange<QualifierRangeFilter> implements Filter, Serializable {
+      extends AbstractByteStringRange<QualifierRangeFilter> implements Filter {
+    private static final long serialVersionUID = -1909319911147913630L;
     private final String family;
 
     private QualifierRangeFilter(String family) {
@@ -505,7 +515,9 @@ public final class Filters {
     }
   }
 
-  public static final class TimestampFilter {
+  public static final class TimestampFilter implements Serializable {
+    private static final long serialVersionUID = 5284219722591464991L;
+
     private TimestampFilter() {}
 
     /**
@@ -529,7 +541,9 @@ public final class Filters {
 
   /** Matches only cells with microsecond timestamps within the given range. */
   public static final class TimestampRangeFilter
-      extends AbstractTimestampRange<TimestampRangeFilter> implements Filter, Serializable {
+      extends AbstractTimestampRange<TimestampRangeFilter> implements Filter {
+    private static final long serialVersionUID = 8410980338603335276L;
+
     private TimestampRangeFilter() {}
 
     @InternalApi
@@ -571,7 +585,9 @@ public final class Filters {
     }
   }
 
-  public static final class ValueFilter {
+  public static final class ValueFilter implements Serializable {
+    private static final long serialVersionUID = 6722715229238811179L;
+
     private ValueFilter() {}
 
     /**
@@ -628,7 +644,9 @@ public final class Filters {
 
   /** Matches only cells with values that fall within the given value range. */
   public static final class ValueRangeFilter extends AbstractByteStringRange<ValueRangeFilter>
-      implements Filter, Serializable {
+      implements Filter {
+    private static final long serialVersionUID = -2452360677825047088L;
+
     private ValueRangeFilter() {}
 
     @InternalApi
@@ -668,7 +686,9 @@ public final class Filters {
     }
   }
 
-  public static final class OffsetFilter {
+  public static final class OffsetFilter implements Serializable {
+    private static final long serialVersionUID = 3228791236971884041L;
+
     private OffsetFilter() {}
 
     /**
@@ -681,7 +701,9 @@ public final class Filters {
     }
   }
 
-  public static final class LimitFilter {
+  public static final class LimitFilter implements Serializable {
+    private static final long serialVersionUID = -794915549003008940L;
+
     private LimitFilter() {}
 
     /**
@@ -705,6 +727,7 @@ public final class Filters {
   }
 
   private static final class SimpleFilter implements Filter {
+    private static final long serialVersionUID = 3595911451325189833L;
     private final RowFilter proto;
 
     private SimpleFilter(@Nonnull RowFilter proto) {
@@ -729,7 +752,7 @@ public final class Filters {
   }
 
   @InternalExtensionOnly
-  public interface Filter extends Cloneable {
+  public interface Filter extends Cloneable, Serializable {
     @InternalApi
     RowFilter toProto();
   }
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Filters.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Filters.java
@@ -26,6 +26,7 @@ import com.google.cloud.bigtable.data.v2.models.Range.AbstractTimestampRange;
 import com.google.common.base.Preconditions;
 import com.google.protobuf.ByteString;
 import java.io.IOException;
+import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import javax.annotation.Nonnull;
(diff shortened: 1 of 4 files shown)


proto-google-cloud-bigtable-v2/src/main/java/com/google/bigtable/v2/RowFilterOrBuilder.java
```diff
@@ -18,10 +18,13 @@
 
 package com.google.bigtable.v2;
 
+import java.io.Serializable;
+
 public interface RowFilterOrBuilder
     extends
-    // @@protoc_insertion_point(interface_extends:google.bigtable.v2.RowFilter)
-    com.google.protobuf.MessageOrBuilder {
+        // @@protoc_insertion_point(interface_extends:google.bigtable.v2.RowFilter)
```
kolea2, line 25: unfortunately anything in the `proto-*` or `grpc-*` modules are auto generated and cannot be modified by hand.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Filters.java
```diff
@@ -200,6 +200,7 @@ public Filter label(@Nonnull String label) {
   // Implementations of target specific filters.
   /** DSL for adding filters to a chain. */
   public static final class ChainFilter implements Filter {
+    private static final long serialVersionUID = -89237431180618430L;
```
igorbernstein2, line 203: Here and everwhere else: The proto builders will need to be marked transient and you will need to implement writeObject/readObject to serialize the built protos:

```
void writeObject(ObjectOutputStream s) {
  s.defaultWriteObject();
  s.writeObject(builder.build());
}
```

proto-google-cloud-bigtable-v2/src/main/java/com/google/bigtable/v2/RowFilter.java
```diff
@@ -18,6 +18,8 @@
 
 package com.google.bigtable.v2;
 
+import java.io.Serializable;
```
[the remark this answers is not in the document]
dmitry-fa: Yes, I see. Is there any way to customize the way how this class is generated?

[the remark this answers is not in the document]
dmitry-fa: Oh, looks like it's not needed.

proto-google-cloud-bigtable-v2/src/main/java/com/google/bigtable/v2/RowFilterOrBuilder.java
dmitry-fa: thanks, the changes are rolled back.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
```diff
@@ -528,4 +540,44 @@ public void labelTest() {
...
+          trySerialization(m.invoke(FILTERS, params));
+        } catch (Exception e) {
+          fail(name + ": " + e);
+        }
+      }
+    }
+  }
+
+  private void trySerialization(Object obj) throws IOException, ClassNotFoundException {
+    Path serFile = Files.createTempFile("filter", ".ser");
```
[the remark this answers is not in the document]
dmitry-fa: done

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
```diff
@@ -528,4 +540,44 @@ public void labelTest() {
...
+    // map methodName -> methodArguments (for those methods which require parameters)
+    Map<String, Object[]> parameterMap = new HashMap<>();
+    parameterMap.put("condition", new Object[] {FILTERS.pass()});
+    parameterMap.put("label", new Object[] {"label"});
+    parameterMap.put("fromProto", new Object[] {FILTERS.label("label").toProto()});
+
+    for (Method m : Filters.class.getDeclaredMethods()) {
+      String name = m.getName();
+      if (Modifier.isPublic(m.getModifiers())) {
+        Object[] params = parameterMap.get(name);
```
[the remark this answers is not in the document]
dmitry-fa: done with a minor change:
```
if (!public) {continue} 
swtich()...
```
->
```
if (public) {
  swtich()...
} 
```

I don't like using `continue`, but if you insist, I'll update as you suggested.

proto-google-cloud-bigtable-v2/src/main/java/com/google/bigtable/v2/RowFilter.java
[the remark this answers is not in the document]
dmitry-fa: changes are rolled back.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
```diff
@@ -528,4 +538,66 @@ public void labelTest() {
...
+            checkSerialization(name, m.invoke(FILTERS));
+        }
+      }
+    }
+  }
+
+  private void checkSerialization(String name, Object filter) {
+    try {
+      Object deserialized = serializeDeserialize(filter);
+      if (filter instanceof Filters.Filter) {
```
igorbernstein2, line 579: why if? when is this false?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
```diff
@@ -528,4 +538,66 @@ public void labelTest() {
...
+        RowFilter protoBefore = ((Filters.Filter) filter).toProto();
+        RowFilter protoAfter = ((Filters.Filter) deserialized).toProto();
+        assertEquals(
```
igorbernstein2, line 582: Please use https://truth.dev/

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
```diff
@@ -528,4 +538,66 @@ public void labelTest() {
...
+            "'" + name + "' filter protoBuf mismatch after deserialization",
+            protoBefore,
+            protoAfter);
+      }
+    } catch (IOException | ClassNotFoundException e) {
+      fail(name + ": " + e);
+    }
+  }
+
+  private Object serializeDeserialize(Object obj) throws IOException, ClassNotFoundException {
```
igorbernstein2, line 592: static?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
```diff
@@ -528,4 +538,66 @@ public void labelTest() {
...
+            checkSerialization(name, FILTERS.label("label").toProto());
+            break;
+          default:
+            checkSerialization(name, m.invoke(FILTERS));
+        }
+      }
+    }
+  }
+
+  private void checkSerialization(String name, Object filter) {
```
igorbernstein2, line 576: static?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
```diff
@@ -528,4 +538,66 @@ public void labelTest() {
...
+      if (filter instanceof Filters.Filter) {
+        RowFilter protoBefore = ((Filters.Filter) filter).toProto();
+        RowFilter protoAfter = ((Filters.Filter) deserialized).toProto();
+        assertEquals(
+            "'" + name + "' filter protoBuf mismatch after deserialization",
+            protoBefore,
+            protoAfter);
+      }
+    } catch (IOException | ClassNotFoundException e) {
+      fail(name + ": " + e);
```
igorbernstein2, line 588: why not just bubble the exception up?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
dmitry-fa: `Filters.Filter` interface declares `toProto()` which I can call to verify the deserialized object.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
dmitry-fa: yes. 
I'm thinking about moving this method to a utility class because it's very generic and could be used by any tests for serialization. Do you know a better place for it?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
dmitry-fa: in case of this exception, one will have a hard time trying to find which filter caused it. `fail(name + ": " + e)` tells the name of filter in trouble.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
dmitry-fa: done

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
dmitry-fa: done

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
dmitry-fa: done

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/models/FiltersTest.java
igorbernstein2: But when is it false? from what I can tell all methods defined on Filters return a Filter. If I'm missing something and there are methods that dont return a Filter, then we need to have an explicit whitelist. As it stands its very easy for this test to miss testing a method

dmitry-fa: it's ~50-50:
a filter 'condition' ConditionFilter
a filter 'label' SimpleFilter
a filter 'chain' ChainFilter
a filter 'interleave' InterleaveFilter
a filter 'pass' SimpleFilter
a filter 'block' SimpleFilter
a filter 'sink' SimpleFilter
not a filter 'family' FamilyFilter
not a filter 'fromProto' RowFilter
not a filter 'qualifier' QualifierFilter
not a filter 'limit' LimitFilter
not a filter 'value' ValueFilter
not a filter 'offset' OffsetFilter
not a filter 'key' KeyFilter
not a filter 'timestamp' TimestampFilter

igorbernstein2: So that means the vast majority of the filters arent getting tested, right?

ie we would want to ensure that qualifier().regex() produces a serializable filter. I think we would want to include a switch case for the not filters and do another loop:


switch(name) {
case "qualifier":
  verifyQualifierFilters();
...
}

verifyQualifierFilters(QualifierFilter f) {
  for (Method m : f.class.getDeclaredMethods()) {
    // ...
  }

dmitry-fa: qualifier().regex()  does not use any object fields it could be made static, so serialization-deserialization doesn't affect it.

I suggested a better approach to verify non-filter instances, please check.

dmitry-fa: @igorbernstein2 please take a look at new tests

dmitry-fa: @igorbernstein2 gentle ping, in case you missed this update of the tests. I did as you asked for.

igorbernstein2 (approved):
lgtm

PR #407, opened 2020-09-03
fix: RowCells are not actually serializeable

### LINKED ISSUE #384: RowCells are not actually serializeable
RowCells might contain UnmodifiableLazyStringList for values, which are not actually serializable. 
We need to add custom hooks to RowCell to fix serialization.

The primary issue is that labels are a UnmodifiableLazyStringList, which is not serializeable

dmitry-fa:
Fixes #384

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowCell.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowCell.java
@@ -20,6 +20,7 @@ import com.google.api.core.InternalExtensionOnly;
 import com.google.auto.value.AutoValue;
 import com.google.cloud.bigtable.data.v2.internal.ByteStringComparator;
 import com.google.common.collect.ComparisonChain;
+import com.google.common.collect.ImmutableList;
 import com.google.protobuf.ByteString;
 import java.io.Serializable;
 import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowCell.java
```diff
@@ -62,7 +63,8 @@ public static RowCell create(
       long timestamp,
       @Nonnull List<String> labels,
       @Nonnull ByteString value) {
-    return new AutoValue_RowCell(family, qualifier, timestamp, value, labels);
+    // Ensure that the list is serializable and optimize for the common case
+    return new AutoValue_RowCell(family, qualifier, timestamp, value, ImmutableList.copyOf(labels));
```
igorbernstein2, line 67: This doesnt optimize for the common case. The common case is empty labels. This is currently creating an extra allocation. Optimization would be:

labels = labels.isEmpty()? ImmutableList.of() : ImmutableList.copyOf(labels);

dmitry-fa: Yes, indeed. I assumed `ImmutableList.of()` accepts var args, but it is not.
Ternary Operator does not work here(the compiler is not happy), so I have to use `if`.

Please take a look.

igorbernstein2 (approved):
lgtm.
Thanks for fixing this

PR #410, opened 2020-09-08
fix: Add documentation to bulkReadRows that each batch will process t…

### LINKED ISSUE #386: Add documentation to bulkReadRows that each batch will process the keys sequentially
ReadRows protocol requires that rows are sent in ascending key order, which implies that each key in the request must be looked up sequentially. We should make this clear in the javadocs to ensure that our customers have the correct expectations

dmitry-fa:
Fixes #386

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -1078,8 +1078,12 @@ public class BigtableDataClient implements AutoCloseable {
 
   /**
    * Reads rows for given tableId in a batch. If the row does not exist, the value will be null. The
-   * keys are processed sequentially one by one in the order they are passed to the Batcher. The
    * returned Batcher instance is not threadsafe, it can only be used from a single thread.
+   *
+   * <p>Performance notice: The ReadRows protocol requires that rows are sent in ascending key
+   * order, which means that the keys are processed sequentially on the server-side, so batching
+   * allows improving throughput but not latency. Low latency might be achieved if send smaller
+   * requests concurrently.
    *
    * <p>Sample Code:
    *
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -1077,9 +1077,9 @@ public void bulkMutateRows(BulkMutation mutation) {
   }
 
   /**
-   * Reads rows for given tableId in a batch. If the row does not exist, the value will be null.
-   * This operation should be called with in a single thread. The returned Batcher instance is not
-   * threadsafe, it can only be used from single thread.
+   * Reads rows for given tableId in a batch. If the row does not exist, the value will be null. The
+   * keys are processed sequentially one by one in the order they are passed to the Batcher. The
+   * returned Batcher instance is not threadsafe, it can only be used from a single thread.
```
igorbernstein2, line 1082: This isnt quite right.
The main issue is that Bigtable ReadRows must return row in lexicographic  row key order. Which means that the keys will be processed sequentially on the serverside. So batching works best when the caller cares more about throughput than latency and the keys are grouped together.

If the caller cares more about latency then sending smaller concurrent request is better

dmitry-fa: Thanks for the clarification, now I see what this issue is about.
Please take a look at the updated variant.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -1113,8 +1117,13 @@ public void bulkMutateRows(BulkMutation mutation) {
...
    * Reads rows for given tableId and filter criteria in a batch. If the row does not exist, the
-   * value will be null. This operation should be called with in a single thread. The returned
-   * Batcher instance is not threadsafe, it can only be used from single thread.
+   * value will be null. The returned Batcher instance is not threadsafe, it can only be used from a
+   * single thread.
+   *
+   * <p>Performance notice: The ReadRows protocol requires that rows are sent in ascending key
+   * order, which means that the keys are processed sequentially on the server-side, so batching
+   * allows improving throughput but not latency. Low latency might be achieved if send smaller
+   * requests concurrently.
```
vermas2012, line 1126: Lower latencies can be achieved by sending smaller requests concurrently.

dmitry-fa: Agree, sounds much better.

igorbernstein2 (approved):
lgtm

PR #413, opened 2020-09-15
chore: update CODEOWNERS

kolea2:


chingor13 (changes requested):
This is controlled by the codeowner_team field in .repo-metadata.json. Can you update it there too?

kolea2:
> This is controlled by the codeowner_team field in .repo-metadata.json. Can you update it there too?

Thanks for catching this, fixed!

chingor13 (approved):

PR #450, opened 2020-10-07
feat: backup level IAM

kolea2:
CC @alexoneill

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClient.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClient.java
@@ -1319,21 +1319,8 @@ public final class BigtableTableAdminClient implements AutoCloseable {
    */
   @SuppressWarnings("WeakerAccess")
   public ApiFuture<Policy> getIamPolicyAsync(String tableId) {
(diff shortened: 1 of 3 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClient.java
```diff
@@ -1463,9 +1434,226 @@ public Policy apply(com.google.iam.v1.Policy proto) {
...
+   *     permissions</a>
+   */
+  @SuppressWarnings({"WeakerAccess"})
+  public ApiFuture<List<String>> testBackupIamPermissionAsync(
+      String clusterId, String backupId, String... permissions) {
+    String backupName = NameUtil.formatBackupName(projectId, instanceId, clusterId, backupId);
+    return testIamPermissions(backupName, permissions);
+  }
+
+  private ApiFuture<Policy> getIamPolicyResult(String name) {
```
igorbernstein2, line 1617: The name sticks out a bit. I'm guessing its to avoid conflict with existing table getIamPolicy.

I think it would be cleaner to name the generic versions as `getResourceIamPolicy`, `setResourceIamPolicy`, etc. or `getIamPolicyBase`

And the public resource specific ones can be called `getBackupIamPolicy`.

igorbernstein2 (approved):
lgtm

PR #451, opened 2020-10-07
chore: make version embedding more robust

igorbernstein2:
The previous approach used the jar manifest. Unfortunately that approaches falls apart when the 
consumer shades the client jar. The new approach uses a combination of releasetool's version tag
replacement and maven-resources-plugin to generate a class with an embedded version string.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/GoogleUtils.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/GoogleUtils.java
@@ -17,8 +17,10 @@ package com.google.cloud.bigtable;
 
 import com.google.api.core.InternalApi;
 
+/** Utility to get the current version of the Bigtable client. */
 @InternalApi("For internal use only")
-public final class Version {
-  // The released version, populated by maven.
-  public static String VERSION="${java-bigtable.version}";
+// Note: this file MUST be named GoogleUtils.java so that it can be picked up by releasetool
+public class GoogleUtils {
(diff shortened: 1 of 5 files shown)


kolea2 (approved):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/GoogleUtils.java
```diff
@@ -0,0 +1,26 @@
...
+package com.google.cloud.bigtable;
+
+import com.google.api.core.InternalApi;
+
+/** Utility to get the current version of the Bigtable client. */
+@InternalApi("For internal use only")
+// Note: this file MUST be named GoogleUtils.java so that it can be picked up by releasetool
+public class GoogleUtils {
+  public static final String BIGTABLE_CLIENT_VERSION =
+      "1.16.1-SNAPSHOT"; // {x-version-update:google-cloud-bigtable:current}
```
kolea2, line 25: @chingor13, to get this to work, I'm guessing we need to add a path here?  https://github.com/googleapis/release-please/blob/master/src/releasers/java-yoshi.ts#L182  Any thing else?

kolea2 (approved):

igorbernstein2:
As discussed offline: This is a temporary solution. In the future we will stop relying on maven resource filtering and instead hardcode the version and bump it by release-please

elharo:
It's worth understanding why the customer was shading, and removing the need for that.

PR #467, opened 2020-10-14
feat: attemp DirectPath by default

WeiranFang:
Update client to attempt DirectPath by default, and remove DirectPath test endpoint.

Note that it doesn't mean that after this change client will just use DirectPath, but will call the DirectPath codepath by default. 

The actually enablement of DirectPath is controlled by service owner via ACL config. For now, after this change, although all users will attempt DirectPath, but they will all just fallback to the original CFE path.

For testing, we will just use `-Dbigtable.directpath-data-endpoint` and `-Dbigtable.directpath-admin-endpoint` to override the default endpoint with DP enabled endpoints.

@igorbernstein2 
@mgarolera
@mohanli-ml

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
@@ -85,9 +85,6 @@ public class EnhancedBigtableStubSettings extends StubSettings<EnhancedBigtableS
   private static final int MAX_MESSAGE_SIZE = 256 * 1024 * 1024;
   private static final String SERVER_DEFAULT_APP_PROFILE_ID = "";
 
-  // TODO(weiranf): Remove this temporary endpoint once DirectPath goes to public beta
-  private static final String DIRECT_PATH_ENDPOINT = "test-bigtable.sandbox.googleapis.com:443";
-
   private static final Set<Code> IDEMPOTENT_RETRY_CODES =
       ImmutableSet.of(Code.DEADLINE_EXCEEDED, Code.UNAVAILABLE);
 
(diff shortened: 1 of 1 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -166,12 +163,6 @@
   private EnhancedBigtableStubSettings(Builder builder) {
     super(builder);
 
-    if (DIRECT_PATH_ENDPOINT.equals(builder.getEndpoint())) {
-      logger.warning(
-          "Connecting to Bigtable using DirectPath."
```
kolea2, line 171: is there anywhere we can keep this logging information to confirm (in tests, for example) that we are connecting using DirectPath?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -245,13 +236,7 @@ public boolean isRefreshingChannel() {
...
             Duration.ofSeconds(10)) // wait this long before considering the connection dead
         .setKeepAliveWithoutCalls(true) // sends ping without active streams
-        // TODO(weiranf): Set this to true by default once DirectPath goes to public beta
-        .setAttemptDirectPath(isDirectPathEnabled());
-  }
-
-  // TODO(weiranf): Remove this once DirectPath goes to public beta
-  private static boolean isDirectPathEnabled() {
-    return Boolean.getBoolean("bigtable.attempt-directpath");
+        .setAttemptDirectPath(true);
```
mgarolera, line 239: Can you add a  comment on why we attemt DirectPath? to give context to future readers.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
igorbernstein2: For tests we inject hooks into netty to ensure that directpath is used, also the integration tests generate verbose grpc logs which would show directpath usage. For prod usage, we dont really have a good way to introspect it as its a grpc level feature

igorbernstein2 (approved):
lgtm

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
WeiranFang: Added comment. PTAL at the wording, thanks!

PR #494, opened 2020-10-23
fix: readRowSettings use manual readRows settings instead of gapic's

noahdietz:
I found that the `readRow` settings were referencing the `baseDefaults` (i.e. the defaults in the GAPIC stub settings) for `readRows`, rather than the manually overridden settings that are used for the `retryableCodes` immediately above. This seems like the safest thing to keep the sources of the RetrySettings all referencing the manually overridden settings.

This will also be necessary for an upcoming config change that will make RPCs configured as non-retryable in the GAPIC to be generated **without** backoff settings (because they are not retryable). @igorbernstein2 this is the change we spoke about.


igorbernstein2 (approved):
lgtm

PR #498, opened 2020-10-26
test(fix): do not skip backup ITs

kolea2:
Cluster ID was null which was causing the tests to skip. Additionally did some cleanup on test structure

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
@@ -17,6 +17,7 @@ package com.google.cloud.bigtable.admin.v2.it;
 
 import static com.google.common.truth.Truth.assertThat;
 import static com.google.common.truth.Truth.assertWithMessage;
+import static com.google.common.truth.TruthJUnit.assume;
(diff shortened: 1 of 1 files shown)


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
```diff
@@ -56,11 +64,11 @@
 
 @RunWith(JUnit4.class)
 public class BigtableBackupIT {
+  @ClassRule public static TestEnvRule testEnvRule = new TestEnvRule();
```
igorbernstein2, line 67: I think this integration test is in a weird state: it was written in fork where TestEnvRule didnt exist and uses its own property names (ie. bigtable.adminendpoint vs bigtable.admin-endpoint. So this test should either migrate to the TestEnvRule now or we should stick to the bespoke property names

kolea2: I'll give migrating a try

kolea2: @igorbernstein2 switched over to TestEnvRule, PTAL

igorbernstein2 (approved):
lgtm

PR #520, opened 2020-11-10
Revert "feat: attemp DirectPath by default"

WeiranFang:
Reverts googleapis/java-bigtable#467

After this PR, gax-java relies on `ComputeEngineCredentials` to determine whether the client should use `ComputeEngineChannelBuilder`: https://github.com/googleapis/gax-java/blob/master/gax-grpc/src/main/java/com/google/api/gax/grpc/InstantiatingGrpcChannelProvider.java#L253

However, `ComputeEngineCredentials` has different assumptions than `ComputeEngineChannelBuilder`, which causes `ComputeEngineChannelBuilder` to fail on AppEngine: https://github.com/grpc/grpc-java/issues/7604.

We need to rollback this PR, and make the fix in gax-java, and then roll-forward again.

@apolcyn 
@mohanli-ml 
@igorbernstein2
@mgarolera


igorbernstein2 (approved):
lgtm

PR #528, opened 2020-11-11
fix: make keepalive params a bit conservative

sushanb:
Having keepalive pings at 10s can cause too_many_ping commands and it will double the keepalive time for new connections on that channel.

This PR sets keepalive pings at 30s (allowed by Google Frontends) which will avoid too many pings warnings.

Do not set keepAliveWithoutCalls(true) as enabling keepalive on idle streams can increase load.


sushanb:
@ejona86 and @igorbernstein2  FYI

kolea2 (approved):

PR #535, opened 2020-11-12
feat: Surface the server-timing metric

mutianf:
- Added a BIGTABLE_GFE_LATENCY_VIEW which will track the GFE t4t7 latency.
- Added a BIGTABLE_GFE_HEADER_MISSING_COUNT_VIEW which will track the number of responses without the server-timing header.
- Added new callables in the callable chains to inject a GrpcMetadataResponse to access the header and trailers. In the callback and response observer, retrieve the serve-timing header from GrpcMetadataResponse, and publish the metrics through HeaderTracer.
- GFE related views can be enabled through a new method BigtableDataSettings.enableGfeOpenCensusStats(). The code will only check and publish the metrics when the GFE views are registered. This is because in the case where direct path is enabled, requests won't go through GFE/CFE and therefore won't have server-timing header.


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableInterceptorProvider.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.google.api.core.BetaApi;
+import com.google.api.gax.grpc.GrpcInterceptorProvider;
+import com.google.cloud.bigtable.data.v2.stub.metrics.ClientHeaderInterceptor;
+import com.google.common.collect.ImmutableList;
+import io.grpc.ClientInterceptor;
+import java.util.List;
+
+@BetaApi
+public class BigtableInterceptorProvider implements GrpcInterceptorProvider {
```
igorbernstein2, line 26: I think it would be better to name this FixedIterceptorProvider and put this in the gaxx package and then in the future to move it to gax proper. (This implies that the construction of the ClientHEaderInterceptor should move out of this class to the caller)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -686,6 +710,29 @@ public EnhancedBigtableStub(EnhancedBigtableStubSettings settings, ClientContext
   }
   // </editor-fold>
 
+  /**
+   * Adds a HeaderTracer instance and current span name to CallOptions so we could surface metrics
+   * in the header with {@link
+   * com.google.cloud.bigtable.data.v2.stub.metrics.ClientHeaderInterceptor}.
+   */
+  private ApiCallContext getContextWithTracer(SpanName spanName) {
```
igorbernstein2, line 718: I think this should be extracted into its own Callable that injects the CallOptions into the passed in ApiCallOptions

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -686,6 +710,29 @@ public EnhancedBigtableStub(EnhancedBigtableStubSettings settings, ClientContext
...
+          "Failed to add tracer in call context. Expected GrpcCallContext but had "
+              + apiCallContext.getClass());
+      return apiCallContext;
+    }
+    GrpcCallContext grpcCallContext = (GrpcCallContext) apiCallContext;
+    CallOptions options = grpcCallContext.getCallOptions();
+    options =
+        options
+            .withOption(HeaderTracer.HEADER_TRACER_CONTEXT_KEY, settings.getHeaderTracer())
+            .withOption(HeaderTracer.SPAN_NAME_CONTEXT_KEY, spanName.toString());
```
igorbernstein2, line 731: I would try to wrap both values in a single object to minimize the overhead of adding CallOptions entries which is backed by an array

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ClientHeaderInterceptor.java
```diff
@@ -0,0 +1,90 @@
...
+              }
+            },
+            headers);
+      }
+    };
+  }
+
+  private void processHeader(Metadata headers, HeaderTracer tracer, String span) {
+    if (tracer == null) {
+      LOGGER.warning("Couldn't find HeaderTracer in call options. Skip exporting gfe metrics");
```
igorbernstein2, line 70: I would add a comment that this shouldn't ever happen because the tracer is injected in ....

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
```diff
@@ -0,0 +1,163 @@
...
+
+  public StatsRecorder getStats() {
+    return stats;
+  }
+
+  public Map<TagKey, TagValue> getStatsAttributes() {
+    return statsAttributes;
+  }
+
+  public void record(MeasureLong measure, long value, @Nullable String span) {
```
igorbernstein2, line 123: I would keep this as a private method and instead expose concrete helpers that describe what you are recording: ie recordGfeLatency(String method, Duration duration)
recordMissingGfeHeader(String method)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
```diff
@@ -0,0 +1,163 @@
...
+import io.opencensus.tags.TagContextBuilder;
+import io.opencensus.tags.TagKey;
+import io.opencensus.tags.TagValue;
+import io.opencensus.tags.Tagger;
+import io.opencensus.tags.Tags;
+import java.util.Map;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+@BetaApi
```
igorbernstein2, line 37: Should this be exposed to users? Unless there is a specific reason to access something, we should limit the exposed surface so that we can evolve our internal apis without worrying about breaking users

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/RpcViews.java
```diff
@@ -31,7 +31,9 @@
           RpcViewConstants.BIGTABLE_COMPLETED_OP_VIEW,
           RpcViewConstants.BIGTABLE_READ_ROWS_FIRST_ROW_LATENCY_VIEW,
           RpcViewConstants.BIGTABLE_ATTEMPT_LATENCY_VIEW,
-          RpcViewConstants.BIGTABLE_ATTEMPTS_PER_OP_VIEW);
+          RpcViewConstants.BIGTABLE_ATTEMPTS_PER_OP_VIEW,
+          RpcViewConstants.BIGTABLE_GFE_LATENCY_VIEW,
```
igorbernstein2, line 35: We should discuss if these should be part of the standard exposed metrics

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -759,6 +772,19 @@ public boolean isRefreshingChannel() {
       return primedTableIds;
     }
 
+    /** Configure the header tracer for surfacing metrics in the header. */
```
igorbernstein2, line 775: We should avoid exposing this to customers. It's an implementation detail for the current approach. When this gets pushed into gax, we will move this into ApiTracer so we should minimize the amount public methods exposed (even if Beta). if this needs to be public then it should be INternalApi

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ClientHeaderInterceptor.java
```diff
@@ -0,0 +1,90 @@
...
+      return;
+    }
+    if (headers.get(SERVER_TIMING_HEADER_KEY) != null) {
+      String serverTiming = headers.get(SERVER_TIMING_HEADER_KEY);
+      Matcher matcher = SERVER_TIMING_HEADER_PATTERN.matcher(serverTiming);
+      tracer.record(RpcMeasureConstants.BIGTABLE_GFE_HEADER_MISSING_COUNT, 0L, span);
```
igorbernstein2, line 76: Do we need to record anything here?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ClientHeaderInterceptor.java
igorbernstein2: I'm also a little concerned about flooding the customer logs here. If the tracer is missing, it is quite likely that it will be missing for all of the requests. So this will log for every request

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ClientHeaderInterceptor.java
```diff
@@ -0,0 +1,90 @@
...
+      if (matcher.find()) {
+        long latency = Long.valueOf(matcher.group("dur"));
+        tracer.record(RpcMeasureConstants.BIGTABLE_GFE_LATENCY, latency, span);
+      } else {
+        LOGGER.warning(
```
igorbernstein2, line 81: Same flooding concern as above

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
```diff
@@ -0,0 +1,163 @@
...
+public class HeaderTracer {
```
igorbernstein2, line 38: Also consider using auto-value

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -127,6 +128,7 @@ public ManagedChannelBuilder apply(ManagedChannelBuilder input) {
                 .setKeepAliveTime(Duration.ofSeconds(30)) // sends ping in this interval
                 .setKeepAliveTimeout(
                     Duration.ofSeconds(10)) // wait this long before considering the connection dead
+                .setInterceptorProvider(BigtableInterceptorProvider.createDefault())
```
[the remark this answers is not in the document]
mutianf: Removed the interceptor, using GrpcResponseMetadata instead.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
mutianf: Updated to InternalApi. This class still needs to be public so EnhancedStubSettings and EnhancedStub can use it.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/RpcViews.java
mutianf: Added a separate function for registering GFE views.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
mutianf: Updated functions to protected.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
mutianf: Refactored HeaderTracer to use autovalue.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ClientHeaderInterceptor.java
mutianf: This logic is moved to HeaderTracer. 
I decided to keep this so if in any case gfe_latency is not recorded (for example, if the regex failed), we can check if gfe metrics were being tracked by check if gfe_header_missing_count recorded any values.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -203,7 +205,25 @@ public static EnhancedBigtableStubSettings finalizeSettings(
...
                 settings.getTracerFactory())));
-
+    builder.setHeaderTracer(
+        builder
+            .getHeaderTracer()
+            .toBuilder()
+            .setStats(stats)
+            .setTagger(tagger)
+            .setStatsAttributes(
+                ImmutableMap.<TagKey, TagValue>builder()
```
igorbernstein2, line 215: Can you make this map instance shared with the attrs above?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+@InternalApi
+@AutoValue
+public abstract class HeaderTracer {
+
+  public static final Metadata.Key<String> SERVER_TIMING_HEADER_KEY =
```
igorbernstein2, line 42: do these have to be public?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
```diff
@@ -0,0 +1,124 @@
...
+  @AutoValue.Builder
+  public abstract static class Builder {
+    // <editor-fold desc="Public API">
+    public abstract Builder setTagger(@Nonnull Tagger tagger);
+
+    public abstract Builder setStats(@Nonnull StatsRecorder stats);
+
+    public abstract Builder setStatsAttributes(@Nonnull Map<TagKey, TagValue> statsAttributes);
+
+    public abstract Tagger getTagger();
```
igorbernstein2, line 55: do we need the getters on the builder?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public abstract Map<TagKey, TagValue> getStatsAttributes();
+
+    abstract HeaderTracer autoBuild();
+
+    public HeaderTracer build() {
+      HeaderTracer headerTracer = autoBuild();
+      Preconditions.checkNotNull(headerTracer.getStats(), "StatsRecorder must be set");
+      Preconditions.checkNotNull(headerTracer.getTagger(), "Tagger must be set");
+      Preconditions.checkNotNull(headerTracer.getStatsAttributes(), "Stats attributes must be set");
```
igorbernstein2, line 67: Unless I;m mistaken, this shouldnt be necessary, autovalue should do null checks for all properties that arent marked Nullable

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracer.java
```diff
@@ -0,0 +1,124 @@
...
+    // </editor-fold>
+  }
+
+  public abstract Tagger getTagger();
+
+  public abstract StatsRecorder getStats();
+
+  public abstract Map<TagKey, TagValue> getStatsAttributes();
+
+  public void recordGfeMetrics(@Nonnull Metadata metadata, String spanName) {
```
igorbernstein2, line 79: Please add a comment as to when spanName might be null. Its a bit surprising to me

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerStreamingCallable.java
```diff
@@ -0,0 +1,106 @@
...
+ * Record GFE metrics.
+ *
+ * <p>This class is considered an internal implementation detail and not meant to be used by
+ * applications.
+ */
+@InternalApi
+public class HeaderTracerStreamingCallable<RequestT, ResponseT>
+    extends ServerStreamingCallable<RequestT, ResponseT> {
+
+  private ServerStreamingCallable<RequestT, ResponseT> innerCallable;
```
igorbernstein2, line 37: I think these should be final

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerStreamingCallable.java
```diff
@@ -0,0 +1,106 @@
...
+  private HeaderTracer headerTracer;
+  private String spanName;
+
+  public HeaderTracerStreamingCallable(
+      @Nonnull ServerStreamingCallable<RequestT, ResponseT> callable,
+      @Nonnull HeaderTracer headerTracer,
+      @Nonnull String spanName) {
+    this.innerCallable = callable;
```
igorbernstein2, line 45: Precondition checks would be good here

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerStreamingCallable.java
```diff
@@ -0,0 +1,106 @@
...
+import io.grpc.Metadata;
+import javax.annotation.Nonnull;
+
+/**
+ * Record GFE metrics.
+ *
+ * <p>This class is considered an internal implementation detail and not meant to be used by
+ * applications.
+ */
+@InternalApi
```
igorbernstein2, line 33: Please add a comment as to what this does. It will be helpful for future maintainers..Something like:

This callable will inject GrpcResponseMetadata into the ApiCallOptions. Upon rpc completion, gax will inject grpc trailers into this class. The trailers will then be processed by HeaderTracer

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerUnaryCallable.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.api.gax.rpc.ApiCallContext;
+import com.google.api.gax.rpc.UnaryCallable;
+import com.google.common.util.concurrent.MoreExecutors;
+import io.grpc.Metadata;
+
+/**
+ * Record GFE metrics.
+ *
+ * <p>This class is considered an internal implementation detail and not meant to be used by
+ * applications.
```
igorbernstein2, line 30: similar to above

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerStreamingCallable.java
igorbernstein2: Also, can you add the new internal api classes to this exclusion list:
https://github.com/googleapis/java-bigtable/blob/master/pom.xml#L252

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/RpcViewConstants.java
```diff
@@ -124,4 +128,22 @@
               BIGTABLE_APP_PROFILE_ID,
               BIGTABLE_OP,
               BIGTABLE_STATUS));
+
+  static final View BIGTABLE_GFE_LATENCY_VIEW =
+      View.create(
+          View.Name.create("cloud.google.com/java/bigtable/gfe_latency"),
+          "GFE t4t7 latency in msecs",
```
igorbernstein2, line 135: This description is meant for end users who don't really know what t4t7 latency means. Can you reword this a bit?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/RpcViewConstants.java
```diff
@@ -124,4 +128,22 @@
...
+          BIGTABLE_GFE_LATENCY,
+          AGGREGATION_WITH_MILLIS_HISTOGRAM,
+          ImmutableList.of(
+              BIGTABLE_INSTANCE_ID, BIGTABLE_PROJECT_ID, BIGTABLE_APP_PROFILE_ID, BIGTABLE_OP));
+
+  static final View BIGTABLE_GFE_HEADER_MISSING_COUNT_VIEW =
+      View.create(
+          View.Name.create("cloud.google.com/java/bigtable/gfe_header_missing_count"),
+          "Number of responses without the server-timing header",
```
igorbernstein2, line 144: Can you add a note what this means? ie. that most likely the request never made it to the google's load balancers?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/RpcViews.java
```diff
@@ -33,15 +33,32 @@
...
+      ImmutableSet.of(
+          RpcViewConstants.BIGTABLE_GFE_LATENCY_VIEW,
+          RpcViewConstants.BIGTABLE_GFE_HEADER_MISSING_COUNT_VIEW);
+
   /** Registers all Bigtable specific views. */
   public static void registerBigtableClientViews() {
     registerBigtableClientViews(Stats.getViewManager());
   }
 
+  /** Register views for GFE metrics. */
```
igorbernstein2, line 46: Please add a lot more docs explaining why you'd wanna do this. Also,`@BetaApi`

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerCallableTest.java
```diff
@@ -0,0 +1,461 @@
...
+  private EnhancedBigtableStub noHeaderStub;
+
+  private static final String PROJECT_ID = "fake-project";
+  private static final String INSTANCE_ID = "fake-instance";
+  private static final String APP_PROFILE_ID = "default";
+  private static final String TABLE_ID = "fake-table";
+
+  private static final long WAIT_FOR_METRICS_TIME_MS = 1_000;
+
+  private int fakeServerTiming;
```
igorbernstein2, line 97: I think this needs to be atomic because this is set by a test thread and read by a server thread. Otherwise the cpu cache line might not be flushed

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerCallableTest.java
```diff
@@ -0,0 +1,461 @@
...
+import org.junit.runners.JUnit4;
+import org.mockito.Answers;
+import org.mockito.Mock;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.stubbing.Answer;
+
+@RunWith(JUnit4.class)
+public class HeaderTracerCallableTest {
```
igorbernstein2, line 77: Would you mind adding tests that ensure that the metric is incremented when server returns a non-ok status as well?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerTest.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Map;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.mockito.Mockito;
+
+@RunWith(JUnit4.class)
+public class HeaderTracerTest {
+
+  private StatsComponent localStats = new StatsComponentImpl();
```
igorbernstein2, line 36: final?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerCallableTest.java
```diff
@@ -0,0 +1,461 @@
...
+  @Rule public MockitoRule mockitoRule = MockitoJUnit.rule();
```
igorbernstein2, line 78: final?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerCallableTest.java
```diff
@@ -0,0 +1,461 @@
...
+                RpcMeasureConstants.BIGTABLE_OP, TagValue.create("Bigtable.ReadRows")),
+            PROJECT_ID,
+            INSTANCE_ID,
+            APP_PROFILE_ID);
+
+    assertThat(mutateRowMissingCount).isEqualTo(mutateRowCalls);
+    assertThat(readRowsMissingCount).isEqualTo(readRowsCalls);
+  }
+
+  private class ReadRowAnswer implements Answer {
```
igorbernstein2, line 381: I think this is a personal taste thing, but why go through mockito for this? Wouldnt it be simpler to just implement a fake BigtableImplBase?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+  @Test
+  public void testEmptyBuilder() {
```
igorbernstein2, line 39: nit. testDefaultBuilder()

igorbernstein2 (approved):
This looks great. I just had a few nit picks

igorbernstein2:
Oh, one more thing: we should chat about DIrectPath interactions

igorbernstein2:
Also please update the readme

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/HeaderTracerStreamingCallable.java
mutianf: Good to know! In this case my classes were added under metrics so I think it's covered by: https://github.com/googleapis/java-bigtable/blob/master/pom.xml#L285

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/RpcViews.java
mutianf: RpcViews class is already marked with @BetaApi so I'm skipping adding it here

igorbernstein2 (approved):
lgtm!

Thanks for putting this together!

PR #586, opened 2021-01-06
fix: Retry "received rst stream"

### LINKED ISSUE #106: Bigtable: INTERNAL: HTTP/2 error code: INTERNAL_ERROR Received Rst Stream as a retryable status code
Long lived connections are sometimes disconnected via a RST frame. These should be handled similar to UNAVAILABLE. Also need to make sure that GOAWAYs are handled similar.

This is the same issue as https://github.com/googleapis/cloud-bigtable-client/issues/2177

mutianf:

Fixes #106  ☕️

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
@@ -346,8 +346,14 @@ public class EnhancedBigtableStub implements AutoCloseable {
                 .build(),
             readRowsSettings.getRetryableCodes());
 
+    // Sometimes ReadRows connections are disconnected via an RST frame. This error is transient and
+    // should be treated similar to UNAVAILABLE. However, this exception has an INTERNAL error code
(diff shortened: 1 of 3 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -365,10 +366,14 @@ public EnhancedBigtableStub(EnhancedBigtableStubSettings settings, ClientContext
         new HeaderTracerStreamingCallable<>(
             watched, settings.getHeaderTracer(), getSpanName("ReadRows").toString());
 
+    // Check for "received rst stream" exceptions and convert them to retryable ApiExceptions
+    ServerStreamingCallable<ReadRowsRequest, RowT> convertException =
+        new ReadRowsConvertExceptionCallable<>(withHeaderTracer);
+
```
igorbernstein2, line 372: I think this should be a lot deeper in the call chain, it modifies the behavior of the exception callable thats created in the raw callable.

mutianf:
@igorbernstein2 Updated the code and comment, ready for another review :)

igorbernstein2 (approved):
looks great, thanks for fixing this!

PR #612, opened 2021-02-04
 feat: dynamic flow control for bulk mutation batcher

mutianf:
Implementation of go/veneer-dynamic-flow-control
Will split this PR into smaller chunks. And it depends on https://github.com/googleapis/gax-java/pull/1288.

Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/java-bigtable/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> ☕️

PR #620, opened 2021-02-08
feat: dynamic flow control part 1

mutianf:
Implementation of go/veneer-dynamic-flow-control.

Part 1, add the settings for enabling / disabling dynamic flow control. The settings will later be used in EnhancedBigtableStub, see https://github.com/googleapis/java-bigtable/pull/612 for reference.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
@@ -220,7 +220,7 @@ public final class BigtableDataSettings {
   }
 
   /**
-   * Gets if latency based throttling is enabled for bulk mutation {@link
+   * Gets if latency based throttling is enabled for {@link
    * BigtableDataClient#newBulkMutationBatcher(String)}
    */
   @BetaApi("Latency based throttling is not currently stable and may change in the future")
@@ -385,9 +385,9 @@ public final class BigtableDataSettings {
     }
 
     /**
-     * Enable latency based throttling for bulk mutation {@link
-     * BigtableDataClient#newBulkMutationBatcher(String)} with a target rpc latency. The number of
-     * allowed in-flight requests will be adjusted to reach the target bulk mutations rpc latency.
+     * Enable latency based throttling for {@link BigtableDataClient#newBulkMutationBatcher(String)}
+     * with a target rpc latency. The number of allowed in-flight requests will be adjusted to reach
+     * the target bulk mutations rpc latency.
      */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
     public Builder enableBatchMutationLatencyBasedThrottling(long targetRpcLatencyMs) {
@@ -396,7 +396,7 @@ public final class BigtableDataSettings {
     }
 
     /**
-     * Disable latency based throttling for bulk mutation {@link
+     * Disable latency based throttling for {@link
      * BigtableDataClient#newBulkMutationBatcher(String)}.
      */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
@@ -406,7 +406,7 @@ public final class BigtableDataSettings {
     }
 
     /**
-     * Gets if latency based throttling is enabled for bulk mutation {@link
+     * Gets if latency based throttling is enabled for {@link
      * BigtableDataClient#newBulkMutationBatcher(String)}
      */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
@@ -391,7 +391,7 @@ public final class BigtableDataSettings {
      */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
     public Builder enableBatchMutationLatencyBasedThrottling(long targetRpcLatencyMs) {
-      stubSettings.enableBatchMutationLatencyBasedThrottling(targetRpcLatencyMs);
+      stubSettings.bulkMutateRowsSettings().setLatencyBasedThrottling(true, targetRpcLatencyMs);
       return this;
     }
 
@@ -401,7 +401,7 @@ public final class BigtableDataSettings {
      */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
     public Builder disableBatchMutationLatencyBasedThrottling() {
-      stubSettings.disableBatchMutationLatencyBasedThrottling();
+      stubSettings.bulkMutateRowsSettings().setLatencyBasedThrottling(false, null);
       return this;
     }
 
@@ -411,7 +411,7 @@ public final class BigtableDataSettings {
      */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
     public boolean isLatencyBasedThrottlingForBatchMutationEnabled() {
-      return stubSettings.isLatencyBasedThrottlingForBatchMutationEnabled();
+      return stubSettings.bulkMutateRowsSettings().isLatencyBasedThrottlingEnabled();
     }
     /**
      * Returns the underlying settings for making RPC calls. The settings should be changed with
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
@@ -229,7 +229,10 @@ public final class BigtableDataSettings {
     return stubSettings.bulkMutateRowsSettings().isLatencyBasedThrottlingEnabled();
   }
 
-  /** Gets target rpc latency if latency based throttling is enabled. Otherwise return null. */
+  /**
+   * Gets target bulk mutation rpc latency if latency based throttling is enabled for {@link
+   * BigtableDataClient#newBulkMutationBatcher(String)}. Otherwise returns null.
+   */
   @BetaApi("Latency based throttling is not currently stable and may change in the future")
   @Nullable
   public Long getBatchMutationsTargetRpcLatencyMs() {
@@ -422,10 +425,13 @@ public final class BigtableDataSettings {
       return stubSettings.bulkMutateRowsSettings().isLatencyBasedThrottlingEnabled();
     }
 
-    /** Gets target rpc latency if latency based throttling is enabled. Otherwise return null. */
+    /**
+     * Gets target bulk mutation rpc latency if latency based throttling is enabled for {@link
+     * BigtableDataClient#newBulkMutationBatcher(String)}. Otherwise returns null.
+     */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
     @Nullable
-    public Long getBatchMutationsTargetRpcLatencyMs() {
+    public Long getTargetRpcLatencyMsForBatchMutation() {
       return stubSettings.bulkMutateRowsSettings().getTargetRpcLatencyMs();
     }
 
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
@@ -406,7 +406,7 @@ public final class BigtableDataSettings {
      */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
     public Builder enableBatchMutationLatencyBasedThrottling(long targetRpcLatencyMs) {
-      stubSettings.bulkMutateRowsSettings().setLatencyBasedThrottling(true, targetRpcLatencyMs);
+      stubSettings.bulkMutateRowsSettings().enableLatencyBasedThrottling(targetRpcLatencyMs);
       return this;
     }
 
@@ -416,7 +416,7 @@ public final class BigtableDataSettings {
      */
     @BetaApi("Latency based throttling is not currently stable and may change in the future")
     public Builder disableBatchMutationLatencyBasedThrottling() {
-      stubSettings.bulkMutateRowsSettings().setLatencyBasedThrottling(false, null);
+      stubSettings.bulkMutateRowsSettings().disableLatencyBasedThrottling();
       return this;
     }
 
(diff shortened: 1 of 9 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -750,6 +751,36 @@ public boolean isRefreshingChannel() {
...
+
+    /**
+     * Gets if latency based throttling is enabled for {@link
+     * BigtableDataClient#newBulkMutationBatcher(String)}
+     */
+    @BetaApi("Latency based throttling is not currently stable and may change in the future")
+    public boolean isLatencyBasedThrottlingForBatchMutationEnabled() {
+      return bulkMutateRowsSettings.isLatencyBasedThrottlingEnabled();
+    }
+
```
igorbernstein2, line 783: I dont think these provide much value, I think we can inline them in BigtableDataSettings

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
```diff
@@ -145,9 +188,92 @@ public Builder setRetrySettings(@Nonnull RetrySettings retrySettings) {
...
+        boolean isLatencyBasedThrottlingEnabled, @Nullable Long targetRpcLatency) {
+      Preconditions.checkState(
+          this.batchingSettings != null,
+          "batchingSettings must be set before enabling / disabling throttling");
+      Preconditions.checkArgument(
+          !isLatencyBasedThrottlingEnabled || targetRpcLatency != null,
+          "target RPC latency must be set if latency based throttling is enabled");
+      Preconditions.checkArgument(
+          targetRpcLatency == null || targetRpcLatency > 0,
+          "if target RPC latency is set, it must be greater than 0");
```
igorbernstein2, line 208: Can this validation move to build() time? This way we wont force our users into set these in a particular order

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
```diff
@@ -145,9 +188,92 @@ public Builder setRetrySettings(@Nonnull RetrySettings retrySettings) {
...
+                .setMaxOutstandingElementCount(maxThrottlingElementCount)
+                .setMinOutstandingElementCount(minElementCount)
+                .setInitialOutstandingRequestBytes(maxThrottlingRequestByteCount)
+                .setMinOutstandingRequestBytes(maxThrottlingRequestByteCount)
+                .setMaxOutstandingRequestBytes(maxThrottlingRequestByteCount)
+                .build();
+        this.flowController = new FlowController(dynamicFlowControlSettings);
+      } else {
+        this.flowController = new FlowController(batchingSettings.getFlowControlSettings());
+      }
```
igorbernstein2, line 246: I think it would be better to defer flowController instantiation to Stub creation. I think that flow control should be scoped to the client instance. ie if do something like
BigtableDataSettings.Builder b = BigtableDataSettings.newBuilder(). enableBatchMutationLatencyBasedThrottling(100);

c1 = BigtableDataClient.create(b.setProjectId("p1").build())
c2 = BigtableDataClient.create(b.setProjectId("p2").build())
I should end up with separate flow controllers

and it will be less surprising that
c2 = BigtableDataClient.create(b.enableBatchMutationLatencyBasedThrottling(100).build())
also has its own flow controller

igorbernstein2 (changes requested):
I wonder if we your PRs should point to a separate branch until they are usable? I'm concerned about having to cut a release of the client with only the public surface exposed w/o any of plumbing

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
mutianf: Moved FlowController creation to EnhancedBigtableStub constructor, and added some tests for EnhancedBigtableStub.create and BigtableDataClientFactory.create

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
mutianf: Removed the checks for BatchingSettings. These 2 are just checking if targetRpcLatency is valid

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -219,6 +219,15 @@ public boolean isRefreshingChannel() {
...
 
+  /**
+   * Gets if latency based throttling is enabled for {@link
+   * BigtableDataClient#newBulkMutationBatcher(String)}
+   */
+  @BetaApi("Latency based throttling is not currently stable and may change in the future")
+  public boolean isLatencyBasedThrottlingForBatchMutationsEnabled() {
+    return stubSettings.bulkMutateRowsSettings().isLatencyBasedThrottlingEnabled();
+  }
+
```
igorbernstein2, line 230: I'm not sure if this getter is necessary, but if you think it is, please be consistent an expose the target latency as well

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
```diff
@@ -375,6 +384,35 @@ public Builder setPrimingTableIds(String... tableIds) {
...
+    }
+
+    /**
+     * Gets if latency based throttling is enabled for {@link
+     * BigtableDataClient#newBulkMutationBatcher(String)}
+     */
+    @BetaApi("Latency based throttling is not currently stable and may change in the future")
+    public boolean isLatencyBasedThrottlingForBatchMutationEnabled() {
+      return stubSettings.bulkMutateRowsSettings().isLatencyBasedThrottlingEnabled();
+    }
```
igorbernstein2, line 415: Same as above, if you expose the getter for one attribute, you should expose the other as well.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -229,6 +235,45 @@ public EnhancedBigtableStub(EnhancedBigtableStubSettings settings, ClientContext
     readModifyWriteRowCallable = createReadModifyWriteRowCallable();
   }
 
+  private FlowController setupBulkMutationFlowController(BigtableBatchingCallSettings settings) {
```
igorbernstein2, line 238: I wonder if we can split this up? I would like to keep the flow controller instantiation here, but would like to keep the default setting computation in a *Settings class. Also if are going to provide default thresholds, then we should set them during instantiation of stub settings so that users can print the defaults and see what they are. Finally the defaults should be documented

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataSettings.java
mutianf: Ok, I'm going to keep the getters here, because all the other settings seem to have getters too?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
```diff
@@ -55,6 +61,9 @@
   // This settings is just a simple wrapper for BatchingCallSettings to allow us to add
   // additional functionality.
   private BatchingCallSettings<RowMutationEntry, Void, BulkMutation, Void> batchingCallSettings;
+  private boolean isLatencyBasedThrottlingEnabled;
+  private Long targetRpcLatencyMs;
+  private DynamicFlowControlSettings dynamicFlowControlSettings;
```
igorbernstein2, line 66: please make these final.
Also if you dont mind, please backfill final for batchingCallSettings

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
```diff
@@ -145,9 +193,106 @@ public Builder setRetrySettings(@Nonnull RetrySettings retrySettings) {
...
+    public DynamicFlowControlSettings getDynamicFlowControlSettings() {
+      return this.dynamicFlowControlSettings;
+    }
+
     /** Builds the {@link BigtableBatchingCallSettings} object with provided configuration. */
     @Override
     public BigtableBatchingCallSettings build() {
+      Preconditions.checkState(batchingSettings != null, "batchingSettings must be set");
+      FlowControlSettings defaultSettings = batchingSettings.getFlowControlSettings();
+      if (isLatencyBasedThrottlingEnabled()) {
```
igorbernstein2, line 254: Please add some comments explaining whats happening here. Something along the lines of:

// Combine static FlowControlSettings with latency based throttling settings to create DynamicFlowControlSettings

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
```diff
@@ -145,9 +193,106 @@ public Builder setRetrySettings(@Nonnull RetrySettings retrySettings) {
...
+        Long maxThrottlingRequestByteCount = defaultSettings.getMaxOutstandingRequestBytes();
+        if (maxThrottlingElementCount == null) {
+          long maxBulkMutateElementPerBatch = 100L;
+          long defaultChannelPoolSize = 2 * Runtime.getRuntime().availableProcessors();
+          maxThrottlingElementCount =
+              Math.min(20_000L, 10L * maxBulkMutateElementPerBatch * defaultChannelPoolSize);
+        }
+        if (maxThrottlingRequestByteCount == null) {
+          maxThrottlingRequestByteCount = 100L * 1024 * 1024;
+        }
```
igorbernstein2, line 265: It seems like you are duplicating the defaults from EnhancedBigtableStubSettings. This will be very hard to keep in sync. If these settings are required and the user hasnt set them, then I think we should raise an error.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
```diff
@@ -145,9 +193,106 @@ public Builder setRetrySettings(@Nonnull RetrySettings retrySettings) {
...
+
+        long initialElementCount = maxThrottlingElementCount / 4;
+        long minElementCount = maxThrottlingElementCount / 100;
```
igorbernstein2, line 268: Please make these numbers a bit less magical. Please explain the rationale for them. Also I think we should start at the max and throttle down if there are performance issues

igorbernstein2 (changes requested):
I like the new approach of always upgrading to DynamicFlowControlSettings, but i think we need to do a bit more work on the defaults

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
```diff
@@ -145,9 +194,135 @@ public Builder setRetrySettings(@Nonnull RetrySettings retrySettings) {
       return this;
     }
 
+    /**
+     * Enables / disables latency based throttling. If enabling the setting, targetRpcLatency needs
+     * to be set.
+     */
+    public Builder setLatencyBasedThrottling(
```
igorbernstein2, line 201: Here and everywhere else `BetaApi` for new public surfaces.

Also, can we split this up into enableLatencyBasedThrottling(targetMs) and disableLatencyBasedThrottling()?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
```diff
@@ -145,9 +194,135 @@ public Builder setRetrySettings(@Nonnull RetrySettings retrySettings) {
...
+    /** Gets if latency based throttling is enabled. */
+    public boolean isLatencyBasedThrottlingEnabled() {
+      return this.isLatencyBasedThrottlingEnabled;
+    }
+
+    /**
+     * Gets the {@link DynamicFlowControlSettings} that'll be used to set up a {@link
+     * FlowController} for throttling.
+     *
+     * <p>By default, this will allow a maximum of 1000 entries per channel of {@link
```
igorbernstein2, line 229: I'm not sure if the defaults should be set/documented here. This class can be used for both reads and writes. I'm not sure that the 2 should share their defaults. Can we document this on the BigtableDataSettings or StubSettings?

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
mutianf: This class already has `BetaApi` annotation, will we still need them for each method?

Will split up the settings.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableBatchingCallSettings.java
mutianf: I thought this class is only for BulkMutation? `public final class BigtableBatchingCallSettings extends UnaryCallSettings<BulkMutation, Void>` and reads uses `BigtableBulkReadRowsCallSettings`.

I put the documentation here because: default `DynamicFlowControlSettings` depends on the default settings for `maxElementCount` and `maxRequestBytes`.  And nowhere else is `DynamicFlowControlSettings` mentioned. So I thought it might be easier to put all the explanations here?

igorbernstein2: naming is hard :(
BigtableBatchingCallSettings  sounds generic and I forgot that we created a separate class for reads

igorbernstein2 (approved):

PR #697, opened 2021-04-01
samples: refactor Bigtable tests using a base class

billyjacobson:


samples/snippets/src/test/java/com/example/bigtable/BigtableBaseTest.java
```diff
@@ -0,0 +1,234 @@
...
+import java.io.IOException;
+import java.io.PrintStream;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Random;
+import java.util.UUID;
+import org.junit.After;
+import org.junit.Before;
+
+public class BigtableBaseTest {
```
kolea2, line 39: is this class actually running any tests? If no, I would rename this as a Helper class, and instead have the tests not extend it but statically import and use these helper methods.

kolea2 (changes requested):

samples/snippets/src/test/java/com/example/bigtable/BigtableBaseTest.java
```diff
@@ -0,0 +1,234 @@
...
+  // Environment variables
+  public static String projectId;
+  public static String instanceId;
+
+  // Test output stream
+  private static PrintStream originalOut;
+  public ByteArrayOutputStream bout;
+
+  // Mobile time series db variables
+  public static final String TABLE_ID = generateTableId("mobile-time-series");
```
kolea2, line 50: I think this class could be broken up a bit. The generic stuff that happens in each test (`@Before` setup, `@After` teardown, etc) should remain here. However, the mobile time series variables and methods are not relevant to every sample and are a bit confusing to have available in things like the channel pooling example. I would take all of that and put it into a static helper class to be called by those tests instead.

kolea2 (changes requested):

samples/snippets/src/test/java/com/example/bigtable/BigtableBaseTest.java
billyjacobson: done

kolea2 (approved):
Looks awesome! Thanks for doing this

PR #712, opened 2021-04-08
test: update zones in cmek test

ad548:
Change-Id: I33491161798816636e3208e14346e880b471b065

We want to work with us-central1-b, us-east1-b & us-east1-c zones specifically for our internal testing. So just changing to a combination that works.
The requisite keys and permissions have been setup already.

### DIFF
--- a/.kokoro/nightly/integration.cfg
+++ b/.kokoro/nightly/integration.cfg
@@ -8,7 +8,7 @@ env_vars: {
 
 env_vars: {
     key: "INTEGRATION_TEST_ARGS"
-    value: "-P bigtable-emulator-it,bigtable-prod-it -Dbigtable.project=gcloud-devel -Dbigtable.instance=google-cloud-bigtable -Dbigtable.table=integration-tests -Dbigtable.kms_key_name=projects/gcloud-devel/locations/us-central1/keyRings/cmek-test-key-ring/cryptoKeys/cmek-test-key -Dbigtable.wait-for-cmek-key-status=true"
+    value: "-P bigtable-emulator-it,bigtable-prod-it -Dbigtable.project=gcloud-devel -Dbigtable.instance=google-cloud-bigtable -Dbigtable.table=integration-tests -Dbigtable.kms_key_name=projects/gcloud-devel/locations/us-east1/keyRings/cmek-test-key-ring/cryptoKeys/cmek-test-key -Dbigtable.wait-for-cmek-key-status=true"
 }
 
 env_vars: {
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableCmekIT.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableCmekIT.java
@@ -122,7 +122,7 @@ public class BigtableCmekIT {
     Cluster cluster = instanceAdmin.getCluster(instanceId, clusterId1);
     assertThat(cluster.getKmsKeyName()).isEqualTo(kmsKeyName);
 
-    String secondZoneId = testEnvRule.env().getPrimaryRegionSecondZone();
+    String secondZoneId = testEnvRule.env().getSecondaryZone();
     instanceAdmin.createCluster(
         CreateClusterRequest.of(instanceId, clusterId2)
             .setZone(secondZoneId)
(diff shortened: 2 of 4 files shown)


.kokoro/nightly/integration.cfg
```diff
@@ -8,7 +8,7 @@ env_vars: {
 
 env_vars: {
     key: "INTEGRATION_TEST_ARGS"
-    value: "-P bigtable-emulator-it,bigtable-prod-it -Dbigtable.project=gcloud-devel -Dbigtable.instance=google-cloud-bigtable -Dbigtable.table=integration-tests -Dbigtable.kms_key_name=projects/gcloud-devel/locations/us-central1/keyRings/cmek-test-key-ring/cryptoKeys/cmek-test-key -Dbigtable.wait-for-cmek-key-status=true"
+    value: "-P bigtable-emulator-it,bigtable-prod-it -Dbigtable.project=gcloud-devel -Dbigtable.instance=google-cloud-bigtable -Dbigtable.table=integration-tests -Dbigtable.kms_key_name=projects/gcloud-devel/locations/us-east1/keyRings/cmek-test-key-ring/cryptoKeys/cmek-test-key -Dbigtable.wait-for-cmek-key-status=true"
```
ad548, line 11: will need @kolea2 to create this key for gcloud-devel project.

kolea2 (changes requested):

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/AbstractTestEnv.java
```diff
@@ -101,17 +101,17 @@ public boolean shouldWaitForCmekKeyStatusUpdate() {
   }
 
   public String getPrimaryZone() {
-    return "us-central1-b";
-  }
-
-  public String getPrimaryRegionSecondZone() {
-    return "us-central1-c";
+    return "us-east1-c";
```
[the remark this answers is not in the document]
ad548: Yea we could do that. I did run tests locally and the other tests weren't affected so felt confident changing this. Either way, it would be good to test with the us-east1 regions as it has 2 zones we can test with - so would be good if you're able to add that key. 
Maybe I can do the dynamic zone determination as another PR so we can unblock our internal stuff?

kolea2 (changes requested):

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/AbstractTestEnv.java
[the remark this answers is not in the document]
ad548: Okay, sounds good. I'll just hard-code change to us-east1 zones 👍

.kokoro/nightly/integration.cfg
kolea2: added and rerunning tests now

kolea2 (approved):

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableCmekIT.java
```diff
@@ -83,7 +83,7 @@ public static void validatePlatform() throws IOException {
     instanceId = AbstractTestEnv.TEST_INSTANCE_PREFIX + Instant.now().getEpochSecond();
     clusterId1 = instanceId + "-c1";
     clusterId2 = instanceId + "-c2";
-    String zoneId = testEnvRule.env().getPrimaryZone();
+    String zoneId = "us-east1-b";
```
igorbernstein2, line 86: I dont want to have resource ids sprinkled in individual tests. AsbtractEnv is intended to hold all environment specific settings. So that if we need to make changes in your test infra, we dont have to grep for all resource identifiers in individual tests.

Easiest way I can think to do this is to expose a method in AbstractTestEnv:
```java
/** Returns test zones that can be used for intra-region testing. This might overlap with primary and secondary zones */
List<String> getMultipleZonesInSameRegion() {
   return ImmutableList.of("us-east1-b", "us-east1-c");
}
*/

And then you can use it as such:
List<String> zones = testEnvRule.env().getMultipleZonesInSameRegion();
List<String> otherZones = Sets.difference(
ImmutableSet.of(testEnvRule.env().getPrimaryRegionSecondZone(), testEnvRule.env().getSecondaryZone()

ImmutableSet.copyOf(zones))

igorbernstein2 (changes requested):

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableCmekIT.java
ad548: Okay, sounds good. I'll make the change now.

igorbernstein2 (approved):
lgtm!

PR #726, opened 2021-04-12
chore: minor test cleanup

igorbernstein2:


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
```diff
@@ -188,7 +188,7 @@ public void createAndGetBackupTest() throws InterruptedException {
           .isEqualTo(expireTime);
       assertWithMessage("Got empty start time in GetBackup API")
           .that(result.getStartTime())
-          .isNotEqualTo(Timestamp.getDefaultInstance());
```
kolea2, line 191: should this check be is not null and not equal to default instance? (what is the default instance?)

igorbernstein2: This is a type mismatch: result.getStartTime() returns a Duration. Which was populated from a Timestamp proto in Backup.fromProto. I think at some point during development, getStartTime returned a Timestamp, which come up as the default instance of the proto. Then at some point before the merge, it was changed to a Duration and when missing it should be null.

kolea2 (approved):

PR #742, opened 2021-04-20
chore: remove duplicate deps (error_prone & autovalue) from bigtable specific pom

igorbernstein2:
google-shared-config already has provisions for autovalue & error_prone. No need for us to track it separately.
This will remove auto-value-annotations from the compile path and into annotation processor path.

### DIFF
--- a/google-cloud-bigtable-deps-bom/pom.xml
+++ b/google-cloud-bigtable-deps-bom/pom.xml
@@ -70,10 +70,6 @@
     </license>
   </licenses>
 
-  <properties>
-    <autovalue.version>1.8</autovalue.version>
-  </properties>
-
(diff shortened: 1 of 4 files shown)


mutianf (commented):
What's the google-cloud-bigtable/EnableAutoValue.txt file for?

igorbernstein2:
> What's the google-cloud-bigtable/EnableAutoValue.txt file for?

Apparently its used to inject AutoValue into the annotation processor?
https://github.com/googleapis/java-shared-config/blob/51c9f68ff1736761b21c921f078ab2c8675ff268/pom.xml#L758

mutianf (approved):

PR #794, opened 2021-05-03
chore: cloud-rad doc generation

eaball35:
Fixes b/179935006

Due to the way v1 html javadocs are currently [configured](https://github.com/googleapis/java-bigtable/blob/master/pom.xml#L206) - we can't produce docs for cloud-rad in exactly the same manner as other java repos. This allows us to generate for cloud-rad (with excluded classes) without changing what we're doing with current doc generation yet.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -286,6 +286,42 @@
         </plugins>
     </build>
 
+    <profiles>
+        <profile>
+            <id>docFX</id>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-javadoc-plugin</artifactId>
+                        <version>3.2.0</version>
+                        <configuration>
+                            <doclet>com.microsoft.doclet.DocFxDoclet</doclet>
+                            <useStandardDocletOptions>false</useStandardDocletOptions>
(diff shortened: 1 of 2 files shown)


kolea2 (commented):
Thanks @eaball35! Can you confirm that running the updated mvn command works and generates properly with the excluded classes?

eaball35:
> Thanks @eaball35! Can you confirm that running the updated mvn command works and generates properly with the excluded classes?

Yes I triggered off this commit and was able to successfully generate cloud-rad docs without the excluded classes included.

kolea2 (approved):

pom.xml
```diff
@@ -286,6 +286,42 @@
...
+                            </additionalOptions>
+                        </configuration>
+                        <executions>
+                            <execution>
+                                <id>aggregate</id>
+                                <goals>
+                                    <goal>aggregate</goal>
+                                </goals>
+                                <configuration>
+                                    <failOnError>false</failOnError>
```
kolea2, line 315: one q - do we want this to fail?

kolea2 (approved):
lgtm with one question

pom.xml
eaball35: Don't need this. I removed it.

PR #801, opened 2021-05-07
chore(test): improve resource handling in tests

igorbernstein2:
* fix resource cleanup: time based resource cleanup expected a prefix followed by a number, but some tests inserted a string in between which would prevent clean. Now all prefixes are generated in a single place
* fix cleanup contention: concurrent tests will try to cleanup resources at the same time. Mitigate this by swallowing not found error
* fix concurrent test runs: the prefix was based on time only, which would causes tests running in the same second to clobber each other
* optimize backup test to only create a single new instance

PR #806, opened 2021-05-13
chore(test): log resource prefix to make debugging easier

igorbernstein2:
Follow up to #801

* move prefix generation to its own instance rule
* log whenever a new prefix generated, which should make it easier to debug test failures

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
@@ -214,7 +214,7 @@ public class BigtableBackupIT {
   @Test
   public void restoreTableTest() throws InterruptedException, ExecutionException {
     String backupId = prefixGenerator.newPrefix();
-    String restoredTableId = prefixGenerator.newPrefix() + "-restore";
+    String restoredTableId = prefixGenerator.newPrefix();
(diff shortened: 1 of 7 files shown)


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
```diff
@@ -210,8 +213,8 @@ public void deleteBackupTest() throws InterruptedException {
 
   @Test
   public void restoreTableTest() throws InterruptedException, ExecutionException {
-    String backupId = testEnvRule.env().newPrefix();
-    String restoredTableId = testEnvRule.env().newPrefix() + "-restore";
+    String backupId = prefixGenerator.newPrefix();
+    String restoredTableId = prefixGenerator.newPrefix() + "-restore";
```
kolea2, line 217: I know this is only in once place (for now), but maybe it could be nicer to do a builder? `newPrefix().append("-restore")`?

igorbernstein2: I actually want to go the other way and remove this usage :)

igorbernstein2: I removed the suffix, ptal

kolea2 (approved):

PR #862, opened 2021-06-04
chore(build): update clirr config using InternalApi annotation

igorbernstein2:
InternalApi classes are not subject to binary compat guarantees. So we should exclude them. This PR adds a small helper that generates exclusions by scanning the classpath. The intention is to run the tool manually and copy/paste the results. Eventually we should figure out how to exclude annotated classes altogether


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/GenerateClirrIgnoredDifferences.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.common.collect.Collections2;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.reflect.ClassPath;
+import com.google.common.reflect.ClassPath.ClassInfo;
+import java.io.IOException;
+import java.util.Collection;
+import org.checkerframework.checker.nullness.compatqual.NullableDecl;
+
+/** Internal helper to update clirr-ignored-differences */
+@InternalApi
```
kolea2, line 31: I don't think this should be included in source code and published. I think https://github.com/googleapis/java-bigtable/tree/master/scripts could be a better spot? Or a util module?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/GenerateClirrIgnoredDifferences.java
```diff
@@ -0,0 +1,82 @@
...
+  private static final Predicate<ClassInfo> IS_PKG_INFO =
+      new Predicate<ClassInfo>() {
+        @Override
+        public boolean apply(@NullableDecl ClassInfo input) {
+          return input.getSimpleName().equals("package-info");
+        }
+      };
+
+  public static void main(String[] args) throws IOException {
+    Collection<ClassInfo> bigtableClasses = getBigtableClasses();
```
kolea2, line 57: This could use some tests asserting things like:
1. an internal class is added to the xml
2. non internal class is not added
3. internal method on a public class (does that exist?) is only ignored at the method level and not class level

kolea2 (changes requested):

igorbernstein2:
I'll close this for now and come back to it when we have a bit more time

PR #966, opened 2021-08-25
build: remove linkage monitor check

kolea2:
See #964 for more details


suztomo (approved):
The owlbot (or synthtool) may reset this change.

Once all repositories remove the check from the required checks, then I'll update the template in synthtool repository.

PR #1004, opened 2021-09-10
fix: parallel execution and verbose grpc logs

### LINKED ISSUE #692: during tests verbose grpc logs are not split by  test

igorbernstein2:
Mucking around with java util logging fails when tests are executed in parallel. 
So make the 2 features exclusive: by default tests are executed in parallel, if verbose logs are needed 
they can be activated by a profile that will disable parallel execution.

Also switch parallel tests to run in threads instead of processes 
and enable parallel execution for unit tests

Fixes #692

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/TestEnvRule.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/TestEnvRule.java
@@ -71,7 +71,7 @@ public class TestEnvRule implements TestRule {
 
   private static final Logger LOGGER = Logger.getLogger(TestEnvRule.class.getName());
   private static final Boolean BIGTABLE_ENABLE_VERBOSE_GRPC_LOGS =
-      Boolean.getBoolean("enable-verbose-grpc-logs");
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/TestEnvRule.java
```diff
@@ -122,9 +125,14 @@ protected void before(Description description) throws Throwable {
   }
 
   private void configureLogging(Description description) throws IOException {
-    if (Strings.isNullOrEmpty(BIGTABLE_GRPC_LOG_DIR)) {
+    if (!BIGTABLE_ENABLE_VERBOSE_GRPC_LOGS) {
       return;
     }
+    Preconditions.checkState(
```
kolea2, line 131: should we be setting this in the kokoro configs?

igorbernstein2: Yes, it will be set in the kokoro directpath tests via the `enable-verbose-grpc-logs` profile

kolea2 (approved):

PR #1007, opened 2021-09-13
chore: migrate to owlbot

Neenu1995:


pom.xml
```diff
@@ -149,7 +149,32 @@
         <dependencies>
             <!-- Production Deps have been moved to google-cloud-bigtable-deps-bom -->
 
-            <!-- Test Deps in alphabetical order -->
+            <dependency>
+        <groupId>com.google.api.grpc</groupId>
```
kolea2, line 153: why are these getting added?

versions.txt
```diff
@@ -6,4 +6,3 @@ grpc-google-cloud-bigtable-admin-v2:2.1.2:2.1.2
 grpc-google-cloud-bigtable-v2:2.1.2:2.1.2
 proto-google-cloud-bigtable-admin-v2:2.1.2:2.1.2
 proto-google-cloud-bigtable-v2:2.1.2:2.1.2
-google-cloud-bigtable-emulator:0.138.2:0.138.2
```
kolea2, line 9: this shouldn't be removed

kolea2 (changes requested):

pom.xml
Neenu1995: Owlbot adds dependencies for all modules in the parent pom. It's one of the features helping it to propose a new client library version without human intervention.

Does this affect the library build?

versions.txt
Neenu1995: Looking into this.

pom.xml
chingor13: This is also just the dependency management section, not the actual dependencies

PR #1029, opened 2021-10-05
fix: fix typo in rst stream message

mutianf:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsConvertExceptionCallable.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsConvertExceptionCallable.java
@@ -77,8 +77,9 @@ public final class ReadRowsConvertExceptionCallable<ReadRowsRequest, RowT>
   private Throwable convertException(Throwable t) {
     // Long lived connections sometimes are disconnected via an RST frame. This error is
     // transient and should be retried.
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsConvertExceptionCallable.java
```diff
@@ -78,7 +78,7 @@ private Throwable convertException(Throwable t) {
     // Long lived connections sometimes are disconnected via an RST frame. This error is
     // transient and should be retried.
     if (t instanceof InternalException) {
-      if (t.getMessage() != null && t.getMessage().contains("Received Rst stream")) {
+      if (t.getMessage() != null && t.getMessage().contains("Received Rst Stream")) {
```
kolea2, line 81: maybe we use a case insensitive comparison?

kolea2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsConvertExceptionCallable.java
igorbernstein2: I would just simplify this to lowercase().contains("rst_stream") || lowercase.contains("rst stream")

kolea2 (approved):

PR #1052, opened 2021-10-28
fix: align ip version preference for the wrapped  emulator

igorbernstein2:
When the  emulator wrapper is started with -Djava.net.preferIPv6Addresses=true on a machine that defaults to ipv4,
the golang emulator and the java client will pick different ip stacks and not be able to connect. This change will
use java to resolve the localhost's ip address and push it down to the golang emulator


kolea2 (dismissed):
Can we add a test with that system property at both true and false to ensure it passes?

igorbernstein2:
We basically have those tests already ... mac os kokoro jobs set the flag and everything else doesnt. And I dont think its worth adding more kokoro jobs for this

google-cloud-bigtable-emulator/src/main/java/com/google/cloud/bigtable/emulator/v2/Emulator.java
```diff
@@ -98,10 +101,24 @@ public synchronized void start() throws IOException, TimeoutException, Interrupt
     }
     this.port = getAvailablePort();
 
+    // Try to align the localhost address across java  & golang emulator
+    // This should fix issues on systems that default  to  ipv4 but the  jvm is started with
```
mutianf, line 105: Just a nit: there are a few extra spaces in these 2 lines of comment?

mutianf (approved):

google-cloud-bigtable-emulator/src/main/java/com/google/cloud/bigtable/emulator/v2/Emulator.java
igorbernstein2: fixed

PR #1077, opened 2021-11-19
fix: add status label to gfe metrics

mutianf:
Refactor `BigtableTracer` and HeaderTracer callables to include status in the label and add Status label to the GFE metric views.


igorbernstein2 (approved):
lgtm!

PR #1082, opened 2021-12-01
fix: remove tracking latency from channel priming requests

diegomez17:
Created a callable that will not publish latency from channel priming to client metrics.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
@@ -132,6 +132,7 @@ public class EnhancedBigtableStub implements AutoCloseable {
 
   private final ServerStreamingCallable<Query, Row> readRowsCallable;
   private final UnaryCallable<Query, Row> readRowCallable;
+  private final UnaryCallable<Query, Row> readRowRawCallable;
   private final UnaryCallable<String, List<KeyOffset>> sampleRowKeysCallable;
   private final UnaryCallable<RowMutation, Void> mutateRowCallable;
   private final UnaryCallable<BulkMutation, Void> bulkMutateRowsCallable;
@@ -267,6 +268,7 @@ public class EnhancedBigtableStub implements AutoCloseable {
 
     readRowsCallable = createReadRowsCallable(new DefaultRowAdapter());
     readRowCallable = createReadRowCallable(new DefaultRowAdapter());
+    readRowRawCallable = createReadRowRawCallable(new DefaultRowAdapter());
     sampleRowKeysCallable = createSampleRowKeysCallable();
     mutateRowCallable = createMutateRowCallable();
     bulkMutateRowsCallable = createBulkMutateRowsCallable();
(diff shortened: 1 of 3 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -327,6 +327,36 @@ public EnhancedBigtableStub(EnhancedBigtableStubSettings settings, ClientContext
...
+   *   <li>Upon receiving the response stream, it will merge the {@link
+   *       com.google.bigtable.v2.ReadRowsResponse.CellChunk}s in logical rows. The actual row
+   *       implementation can be configured in by the {@code rowAdapter} parameter.
+   *   <li>Retry/resume on failure.
+   *   <li>Filter out marker rows.
+   * </ul>
+   *
+   * <p>NOTE: the caller is responsible for adding tracing & metrics.
+   */
+  protected <RowT> UnaryCallable<Query, RowT> createReadRowRawCallable(
```
mutianf, line 345: I wonder if we could just reuse `#createReadRowsRawCallable`: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java#L294

So in `BigtableChannelPrimer` we could do something like:
```
ApiFuture<Row> f =
     stub.createReadRowsRawCallable(new DefaultRowAdapter())
        .first()
        .futureCall(
            ReadRowsRequest
                 .newBuilder()
                 .setTableName(TableName.format(primingSettings.getProjectId(), primingSettings.getInstanceId(), tableId))
                       ...)
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -327,6 +327,36 @@ public EnhancedBigtableStub(EnhancedBigtableStubSettings settings, ClientContext
...
+      RowAdapter<RowT> rowAdapter) {
+    ServerStreamingCallable<ReadRowsRequest, RowT> readRowsCallable =
+        createReadRowsBaseCallable(
```
mutianf, line 348: I think here we could do `createReadRowsBaseCallable(settings.readRowsSettings(), rowAdapter)` instead of recreating the settings

mutianf (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
diegomez17: I removed this method in favor of the previous comment, it removes the need for this method.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
diegomez17: Oh that is a great way to reduce adding more callables to the Stub, done!

mutianf (approved):

PR #1131, opened 2022-01-11
deps: Testing overriding protobuf version to 3.16.1

prash-mi:
Testing overriding protobuf version to 3.16.1. Not to be merged.

### DIFF
--- a/google-cloud-bigtable-emulator/pom.xml
+++ b/google-cloud-bigtable-emulator/pom.xml
@@ -128,6 +128,12 @@
 
   <dependencyManagement>
     <dependencies>
+      <!-- override protobuf-java version -->
+      <dependency>
+        <groupId>com.google.protobuf</groupId>
+        <artifactId>protobuf-java</artifactId>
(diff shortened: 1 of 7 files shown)


google-cloud-bigtable-emulator/pom.xml
```diff
@@ -128,6 +128,12 @@
...
     <dependencies>
+      <!-- override protobuf-java version -->
+      <dependency>
+        <groupId>com.google.protobuf</groupId>
+        <artifactId>protobuf-java</artifactId>
+        <version>${protobuf-java-test.version}</version>
+      </dependency>
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-bigtable-deps-bom</artifactId>
```
kolea2, line 139: same comment - I think adding this change just to the deps-bom should be sufficient as it should pull the version number from there.

prash-mi: Sure, that makes sense. I will update the BOM when a release is required

PR #1157, opened 2022-02-04
chore(test): fix api breakage in DirectPath integration tests

kongweihan:
The ComputeEngineChannelBuilder layer no longer exists and thus should be removed.

This can only be tested by running kokoro test on GCE. Local IT tests pass but can't prove that this is going to fix it in kokoro.


igorbernstein2:
Thanks for fixing this!
Can you fix formatting by running `mvn com.coveo:fmt-maven-plugin:format`

igorbernstein2:
I ran this on a gce vm and can confirm that this works

igorbernstein2 (approved):
lgtm

Issue #1192, opened 2022-03-04
Ability to start bigtable-emulator on fixed port-number


rameshj28787:
com.google.cloud.bigtable.emulator.v2.Emulator creates the bigtable-emulator in random port-numbers. 

Having the port-number generated __after__ test suite starts execution makes it troublesome to use bigtable-emulator for testing application code.

Because now, we need to dynamically setup BIGTABLE_EMULATOR_HOST environment variable to make application-code connect to emulator. 

Dynamically updating environment variables is not an easy task and not supported by most frameworks (Springboot, Microanut included). There may be external libraries like System Rules, System Lambda etc. to dynamically update environment variables. 
(Reference Link)[https://www.inovex.de/de/blog/cloud-bigtable-unit-testing-java/]  But they could be often not easy to use or not production-ready.

Hence, To use bigtable-emulator for unit-testing or integration testing application code, it will be very helpful & convenient, if we can start bigtable-emulator on a predefined fixed port-number.  (that way we can setup BIGTABLE_EMULATOR_HOST environment variable in Maven/Gradle configurations)

com.google.cloud.bigtable.emulator.v2.Emulator could support creating emulator instances with a predefined port-number 

Currently, the workaround I am using to define a custom FixedPortEmulator class in application testing code that takes and uses port number from BIGTABLE_EMULATOR_HOST environment variable.

rameshj28787:
Pull request: https://github.com/googleapis/java-bigtable/pull/1193

igorbernstein2:
I dont think this usecase make sense here. The primary purpose of the Emulator wrapper is to start using java apis. If there is a need to start an emulator outside of the jvm, then the underlying cbtemulator binary should be used. That binary takes a port as an argument

rameshj28787:
Hi @igorbernstein2 
For our use-case, the emulator is running inside a JVM. We run integration tests with Maven (which is a Java tool) Failsafe plugin. When our integration test-suite executes, it starts our application on a fixed-port (8085) and the app connects to BT-emulator running on another fixed port (8086). We could achieved it only by duplicating Emulator class and making Pull request #1193 changes on our new class.

Is there a better way to do this? Can you please share any examples?

PR #1193, opened 2022-03-04
feat: Add ability to start bigtable emulator on fixed port

rameshj28787:
Change Summary: 

fixes: #1192
<br>
Read port number from BIGTABLE_EMULATOR_HOST environment variable 
- If available, use it for creating emulator instance
- else use 0 (random port numbers)


igorbernstein2 (approved):

igorbernstein2:
I dont think this usecase make sense here. The primary purpose of the Emulator wrapper is to start using java apis. If there is a need to start an emulator outside of the jvm, then the underlying cbtemulator binary should be used. That binary takes a port as an argument

PR #1234, opened 2022-04-21
fix(java): enable skipped tests for Native Image testing 

mpeddada1:
Few changes to note:
- Added gax-grpc to emulator module to bring in Native Image netty configurations
- Ignoring tests that use Mockito for native image testing
- Added resource configuration


mpeddada1:
According to the logs:
```
 5 user-provided feature(s)
  - com.google.cloud.nativeimage.features.clients.CloudFunctionsFeature
  - com.google.cloud.nativeimage.features.clients.CloudSqlFeature
  - com.google.cloud.nativeimage.features.clients.SpannerFeature
  - com.oracle.svm.thirdparty.gson.GsonFeature
  - org.graalvm.junit.platform.JUnitPlatformFeature
```
The features in Gax are not getting  picked up

mpeddada1:
Native Image Tests are now passing.

suztomo (approved):

google-cloud-bigtable-emulator/pom.xml
```diff
@@ -147,6 +148,14 @@
...
     </dependency>
 
+  <!-- This is only needed when we're running tests in native image mode-->
+    <dependency>
+      <groupId>com.google.api</groupId>
+      <artifactId>gax-grpc</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <!-- This is only needed when we're running tests in native image mode-->
+
```
igorbernstein2, line 158: Should this be put in a separate graalvm specific profile ?

google-cloud-bigtable-emulator/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-bigtable-emulator/resource-config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "resources":{
+    "includes":[
+      {
+        "pattern":"\\Qgcloud/bigtable-linux-x86_64/platform/bigtable-emulator/cbtemulator\\E"
+      }
+    ]},
+  "bundles":[]
+}
```
igorbernstein2, line 9: Not sure how graalvm works, but it looks like this will restrict this wrapper to just x64 linux. Currently it supports {darwin,linux,windows}_{x64,aarch64}.

google-cloud-bigtable-emulator/src/test/java/com/google/cloud/bigtable/emulator/v2/ITBigtableEmulatorRuleTest.java
```diff
@@ -38,7 +38,7 @@
 import org.junit.runners.JUnit4;
 
 @RunWith(JUnit4.class)
-public class BigtableEmulatorRuleTest {
+public class ITBigtableEmulatorRuleTest {
```
igorbernstein2, line 41: Why does this need to be changed into an integration test?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientTests.java
```diff
@@ -81,7 +81,7 @@
 import org.mockito.stubbing.Answer;
 
 @RunWith(JUnit4.class)
-public class BigtableInstanceAdminClientTest {
+public class BigtableInstanceAdminClientTests {
```
igorbernstein2, line 84: Why does this need to be renamed? Same for the other pluralizations

igorbernstein2 (changes requested):

google-cloud-bigtable-emulator/pom.xml
suztomo: Memo:

```
~/cloud-opensource-java $ jar tf ~/Downloads/gax-grpc-2.16.0.jar |grep -i netty                      git[branch:gcb_0.173.0]
com/google/api/gax/grpc/nativeimage/GrpcNettyFeature.class
com/google/api/gax/grpc/nativeimage/substitutions/NettyInternalLoggerFactorySubstitutions$OnlyIfInClassPath.class
com/google/api/gax/grpc/nativeimage/substitutions/NettyInternalLoggerFactorySubstitutions.class
```

google-cloud-bigtable-emulator/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-bigtable-emulator/resource-config.json
suztomo: Memo:

```
suztomo@suztomo-macbookpro.roam.internal ~/cloud-opensource-java
~/cloud-opensource-java $ jar tf ~/Downloads/google-cloud-bigtable-emulator-0.143.2.jar |grep cbtemulator
gcloud/bigtable-darwin-arm/platform/bigtable-emulator/cbtemulator
gcloud/bigtable-linux-x86_64/platform/bigtable-emulator/cbtemulator
gcloud/bigtable-darwin-x86_64/platform/bigtable-emulator/cbtemulator
gcloud/bigtable-linux-x86/platform/bigtable-emulator/cbtemulator
gcloud/bigtable-windows-x86_64/platform/bigtable-emulator/cbtemulator.exe
gcloud/bigtable-linux-arm/platform/bigtable-emulator/cbtemulator
gcloud/bigtable-windows-x86/platform/bigtable-emulator/cbtemulator.exe
```

google-cloud-bigtable-emulator/pom.xml
mpeddada1: Thanks for pointing this out @igorbernstein2. Just double checked this and it turns out that these native image configurations (as listed by @suztomo) from gax are needed by the actual application class, `BigtableEmulatorRule`. We need this to be shipped as part of the library. Here is the stacktrace that indicated this:
 
```
  JUnit Vintage:ITBigtableEmulatorRuleTest:testTableAdminClient
    MethodSource [className = 'com.google.cloud.bigtable.emulator.v2.ITBigtableEmulatorRuleTest', methodName = 'testTableAdminClient', methodParameterTypes = '']
    => java.lang.NoClassDefFoundError: Could not initialize class io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder
       io.grpc.netty.shaded.io.grpc.netty.NettyChannelProvider.builderForAddress(NettyChannelProvider.java:38)
       io.grpc.netty.shaded.io.grpc.netty.NettyChannelProvider.builderForAddress(NettyChannelProvider.java:24)
       io.grpc.ManagedChannelBuilder.forAddress(ManagedChannelBuilder.java:39)
       com.google.cloud.bigtable.emulator.v2.Emulator.newChannelBuilder(Emulator.java:289)
       com.google.cloud.bigtable.emulator.v2.Emulator.getAdminChannel(Emulator.java:216)
       com.google.cloud.bigtable.emulator.v2.BigtableEmulatorRule.getAdminChannel(BigtableEmulatorRule.java:78)
       com.google.cloud.bigtable.emulator.v2.ITBigtableEmulatorRuleTest.setUp(ITBigtableEmulatorRuleTest.java:48)
       java.lang.reflect.Method.invoke(Method.java:566)
       org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
       org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
       [...]
```

google-cloud-bigtable-emulator/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-bigtable-emulator/resource-config.json
mpeddada1: Thank you both for the insights! I used a tool called the [trace-agent](https://www.graalvm.org/22.0/reference-manual/native-image/Agent/) which provides hints on the configurations needed and it looks like the suggestions it provided were restricted to just my machine's architecture. I have [added the wrappers for the architectures supported by GraalVM](https://www.graalvm.org/22.0/docs/introduction/#:~:text=GraalVM%20is%20available%20for%20Linux,on%20ARM%2064%2Dbit%20system).

google-cloud-bigtable-emulator/src/test/java/com/google/cloud/bigtable/emulator/v2/ITBigtableEmulatorRuleTest.java
mpeddada1: The native profile in java-shared-config (as [seen here](https://github.com/googleapis/java-shared-config/blob/b5e05c6726c60a7ad7c0d9588b4fab69a0b4e6b7/pom.xml#L811)) enables tests for native image testing based on the naming conventions: `IT*` and `*ClientTest`. This is why we needed to rename this to be picked up by the profile.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientTests.java
mpeddada1: Graalvm at the moment is incompatible with EasyMock and Mockito. We had to rename this class to exclude it from native image testing. As mentioned above, we use this naming convention to enable native image testing for tests: `IT*` and `*ClientTest`

google-cloud-bigtable-emulator/src/main/resources/META-INF/native-image/com.google.cloud/google-cloud-bigtable-emulator/resource-config.json
suztomo: My thinking yesterday was that google-cloud-bigtable-emulator is not for users' deployed application and thus users don't expect it to work in native-image.

Today's thinking is that if a user writes a Java test with the emulator and wants to confirm the test runs on native-image, it's helpful for them if this google-cloud-bigtable-emulator module works out-of-the-box.

google-cloud-bigtable-emulator/pom.xml
igorbernstein2: I'm very confused: test deps dont get shipped (ie they arent transitive), they are only used when executing tests in current module. So if you want them shipped, you need to mark them as runtime scope

google-cloud-bigtable-emulator/src/test/java/com/google/cloud/bigtable/emulator/v2/ITBigtableEmulatorRuleTest.java
igorbernstein2: doesnt this mean that we need to change all of tests in this entire project for them to be run by graal? Also, this needs to be documented so that when we add tests we use the correct naming scheme

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientTests.java
igorbernstein2: As above, this needs to be documented and all of the classes that dont aren't compatible need to have an explicit marker that they are not run by graalvm. Pluralizing the classname is too subtle

mpeddada1: Thanks for the suggestion @igorbernstein2! I agree that the class name change is a little too subtle. I have added a javadoc detailing the exclusion.

suztomo: @Neenu1995 Didn't we have a guide about naming convention for integration tests for client libraries?

google-cloud-bigtable-emulator/pom.xml
mpeddada1: Yes, that is a very valid thought! I think this dependency has already been moved out of `test` scope into `compile` scope. Ahead-of-time compilation allows one to compile their application into a lightweight native executable during compile time. During this process, the compiler does a static analysis of all the reachable code in app so this is really when we would need the configurations that are brought in by gax-grpc.

mpeddada1:
@igorbernstein2 thanks again for your suggestions! any change you could give this another review?

igorbernstein2 (approved):
lgtm!
thanks for putting this together

igorbernstein2 (changes requested):
Oops, noticed one more thing: is there any way to remove the gax dependency?

mpeddada1:
> Oops, noticed one more thing: is there any way to remove the gax dependency?

Mostly summarizing the offline discussion: using a `provided` scope only allows for the configurations to be included in the test jar. As mentioned [here](https://www.graalvm.org/22.1/reference-manual/native-image/BuildConfiguration/#embedding-a-configuration-file), the native image builder expects the configurations to be embedded into the project JAR through the `META-INF/native-image/native-image.properties` or `META-INF/native-image/reflect-config.json` and since   `com.google.cloud.bigtable.emulator.v2.Emulator` references some netty classes, we still need these configurations in the main project JAR.

PR #1258, opened 2022-05-20
build(fix): add in IT args for graalvm presubmit

kolea2:


mpeddada1:
Thanks for fixing this @kolea2! qq, do these arguments apply to only Java 17 tests?

mpeddada1 (approved):

kolea2:
> Thanks for fixing this @kolea2! qq, do these arguments apply to only Java 17 tests?

good catch, they should be fixed in both. I will update.

mpeddada1 (approved):
Looks great!

PR #1288, opened 2022-06-27
chore: mark native image checks as required

mpeddada1:


mpeddada1:
The native image tests have recently started failing with ` Test configuration file wasn't found. Make sure that test execution wasn't skipped.` for the `google-cloud-bigtable-stats` module.  This usually happens with the native maven plugin  when it can't find tests to ru. Note that tests are only enabled for native image testing if they follow the convention: `IT*` or `*ClientTest`. This is a limitation of the plugin and seems to be addressed in the latest [version](https://github.com/graalvm/native-build-tools/releases/tag/0.9.12). We can test this out once java-shared-config is [released and updated](https://github.com/googleapis/java-shared-config/pull/484).

PR #1311, opened 2022-07-14
fix: enable integration test for google-cloud-bigtable-stats

mutianf:


google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/StatsWrapper.java
```diff
@@ -49,4 +51,16 @@ public static List<String> getOperationLatencyViewTagValueStrings() {
         .map(x -> x.asString())
         .collect(Collectors.toCollection(ArrayList::new));
   }
+
+  // A workaround to run BuiltinViewConstantsTest as integration test
```
igorbernstein2, line 55: Please be more descriptive. Explain that the main classes will be relocated when they are packaged but the integration test files will not be. So the integration tests can't reference any transitive deps that have been relocated.

igorbernstein2 (approved):

PR #1317, opened 2022-07-19
feat: add storage utilization gib per node for autoscaling

kolea2:


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableInstanceAdminClientIT.java
```diff
@@ -343,6 +345,7 @@ public void createClusterWithAutoscalingAndPartialUpdateTest() {
                   ClusterAutoscalingConfig.of("ignored", clusterId)
                       .setMaxNodes(4)
                       .setMinNodes(1)
+                      .setStorageUtilizationGibPerNode(2561)
```
kongweihan, line 348: Should we assert this number somewhere below?

kongweihan (approved):
I'm not very familiar with the java client code, but the changes themselves looks reasonable and good!
Had 1 comment in the test.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableInstanceAdminClientIT.java
kongweihan: Oh I see you added it in another commit.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ClusterAutoscalingConfig.java
```diff
@@ -104,6 +104,21 @@ public ClusterAutoscalingConfig setCpuUtilizationTargetPercent(int cpuUtilizatio
...
+    builder.setUpdateMask(
+        FieldMaskUtil.union(
+            builder.getUpdateMask(),
+            FieldMaskUtil.fromString(
+                Cluster.class,
+                "cluster_config.cluster_autoscaling_config.autoscaling_targets.storage_utilization_gib_per_node")));
+    clusterConfigBuilder
+        .getClusterAutoscalingConfigBuilder()
+        .getAutoscalingTargetsBuilder()
+        .setStorageUtilizationGibPerNode(storageUtilizationGibPerNode);
```
mutianf, line 118: Do we need to check if this is a valid value?

kolea2: The server will give back an `INVALID_ARGUMENT` error if the value is not within the range specified in the comments.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CreateClusterRequest.java
```diff
@@ -116,6 +116,7 @@ public CreateClusterRequest setScalingMode(@Nonnull ClusterAutoscalingConfig aut
     int minNodes = autoscalingConfig.getMinNodes();
     int maxNodes = autoscalingConfig.getMaxNodes();
     int cpuTargetPercent = autoscalingConfig.getCpuUtilizationTargetPercent();
+    int storageUtilizationGibPerNode = autoscalingConfig.getStorageUtilizationGibPerNode();
```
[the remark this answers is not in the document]
kolea2: good catch, done!

mutianf (approved):

PR #1336, opened 2022-08-02
deps: upgrade shared config to 1.5.3, exclude google-http-client and google-http-client-gson from gax in google-cloud-bigtable-stats

blakeli0:


google-cloud-bigtable-stats/pom.xml
```diff
@@ -32,6 +32,16 @@
         <dependency>
             <groupId>com.google.api</groupId>
             <artifactId>gax</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.google.http-client</groupId>
+                    <artifactId>google-http-client</artifactId>
```
suztomo, line 38: Are you sure that GAX can work without google-http-client, in library users' environment?

blakeli0: This is only excluding `google-http-client` from `google-cloud-bigtable-stats`, not from `google-cloud-bigtable`.  I checked the usage of gax in `google-cloud-bigtable-stats`, which appears to be only referencing a few annotations, plus the integration tests should catch it if they are required. 
We can make all checks pass by either excluding these here or add them to the exclude list in the dependency list check [here](https://github.com/googleapis/java-bigtable/blob/cb539b50d98ec2a8538ce4691b2639426ca95464/.kokoro/dependencies.sh#L69).
Is there any best practices that prefer not to exclude dependencies? If not, I feel this solution is better since we don't need them, so instead of exclude them in the check, I'd rather exclude them in the pom directly.

suztomo: Thank you. Excluding the unnecessary dependency is better in that has difference in users class path. It's good that these dependencies are not used.

suztomo (approved):

kolea2 (approved):
thank you!

PR #1344, opened 2022-08-08
feat: Implement ReadChangeStreamResumptionStrategy

tengzhonger:


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamResumptionStrategy.java
```diff
@@ -0,0 +1,96 @@
...
+  @Override
+  public ReadChangeStreamRequest getResumeRequest(ReadChangeStreamRequest originalRequest) {
+    // A null token means that we have not successfully read a Heartbeat nor a ChangeStreamMutation,
+    // so start from the beginning.
+    if (this.token == null) {
+      return originalRequest;
+    }
+
+    Builder builder = originalRequest.toBuilder();
+    // We need to clear either start_time or continuation_tokens.
```
mutianf, line 82: Should it be
```suggestion
    // We need to clear start_time and continuation_tokens.
```
? Looks like both start from and continuation tokens are updated

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamResumptionStrategy.java
```diff
@@ -0,0 +1,96 @@
...
+    // And just use the StreamPartition and the token to resume the request.
+    builder.clearStartFrom();
+    builder.setContinuationTokens(
+        StreamContinuationTokens.newBuilder()
+            .addTokens(
+                StreamContinuationToken.newBuilder()
+                    .setPartition(originalRequest.getPartition())
```
mutianf, line 89: Will partition always be the same as the originalRequest?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
```diff
@@ -0,0 +1,494 @@
...
+    // Resume the request with the token from the Heartbeat. `startTime` is cleared.
+    // We don't care about the response here so just do expectRequest.
+    service.expectations.add(
+        RpcExpectation.create()
+            .expectRequest(
+                StreamContinuationTokens.newBuilder().addTokens(streamContinuationToken).build()));
+    List<ChangeStreamRecord> actualResults = getResults();
+    // This is the Heartbeat we get before UNAVAILABLE.
+    Assert.assertEquals(actualResults.size(), 1);
+    Assert.assertTrue(actualResults.get(0) instanceof Heartbeat);
```
mutianf, line 271: Should we check the token here?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
```diff
@@ -0,0 +1,494 @@
...
+  public void errorAfterSingleCloseStreamShouldResumeWithOriginalRequestTest() {
+    // CloseStream.
+    ReadChangeStreamResponse closeStreamResponse =
+        ReadChangeStreamResponse.newBuilder().setCloseStream(createCloseStream()).build();
+    service.expectations.add(
+        RpcExpectation.create()
+            .expectInitialRequest()
+            .respondWith(closeStreamResponse)
+            .respondWithStatus(Code.UNAVAILABLE));
+    // Resume the request with the original request.
```
mutianf, line 285: Should we resume the request in this case if we already see close stream response? Will we get same response if we send the original request?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
```diff
@@ -0,0 +1,494 @@
...
+    // We don't care about the response here so just do expectRequest.
+    service.expectations.add(
+        RpcExpectation.create()
+            .expectRequest(
+                StreamContinuationTokens.newBuilder()
+                    .addTokens(createStreamContinuationToken(DATA_CHANGE_TOKEN))
+                    .build()));
+    List<ChangeStreamRecord> actualResults = getResults();
+    Assert.assertEquals(actualResults.size(), 1);
+    Assert.assertTrue(actualResults.get(0) instanceof ChangeStreamMutation);
```
mutianf, line 314: should we check the token?

mutianf (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamResumptionStrategy.java
tengzhonger: Yes. Updated the comment.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
tengzhonger: I don't think we need to since the `expectRequest` above already checks that the resumption request uses the correct token?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
tengzhonger: Ditto.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
```diff
@@ -282,10 +285,10 @@ public void errorAfterSingleCloseStreamShouldResumeWithOriginalRequestTest() {
             .expectInitialRequest()
             .respondWith(closeStreamResponse)
             .respondWithStatus(Code.UNAVAILABLE));
-    // Resume the request with the original request.
-    // We don't care about the response here so just do expectRequest.
-    service.expectations.add(RpcExpectation.create().expectInitialRequest());
+    // Request is not resumed. We'll get a CloseStream and then an UNAVAILABLE error.
     List<ChangeStreamRecord> actualResults = getResults();
+    expect.expect(UnavailableException.class);
```
mutianf, line 290: Are we also returning OK status in this case? I think it make sense to return OK status since it's already the end of the current stream.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamResumptionStrategy.java
```diff
@@ -0,0 +1,105 @@
...
+    // so start from the beginning.
+    if (this.token == null) {
+      return originalRequest;
+    }
+
+    Builder builder = originalRequest.toBuilder();
+    // We need to clear both start_time and continuation_tokens.
+    // And just use the StreamPartition and the token to resume the request.
+    // The partition should always be the same as the one from the original request,
+    // because otherwise we would have received a CloseStream with different
```
mutianf, line 91:
```suggestion
    // otherwise we would receive a CloseStream with different
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamResumptionStrategy.java
```diff
@@ -0,0 +1,105 @@
...
+  @Override
+  public ReadChangeStreamRequest getResumeRequest(ReadChangeStreamRequest originalRequest) {
+    // A null token means that we have not successfully read a Heartbeat nor a ChangeStreamMutation,
+    // so start from the beginning.
+    if (this.token == null) {
+      return originalRequest;
+    }
+
+    Builder builder = originalRequest.toBuilder();
+    // We need to clear both start_time and continuation_tokens.
```
mutianf, line 88:
```suggestion
    // We need to clear the start_time and 
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamResumptionStrategy.java
```diff
@@ -0,0 +1,105 @@
...
+    // And just use the StreamPartition and the token to resume the request.
```
mutianf, line 89:
```suggestion
    // use the updated continuation_tokens to resume the request.
```

mutianf (approved):
lgtm after the nit

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
tengzhonger: Discussed offline. Let's don't worry about "UNAVAILABLE after CloseStream" since the server will return an OK right after sending a CloseStream.

PR #1347, opened 2022-08-11
feat!: rename ListChangeStreamPartitions to GenerateInitialChangeStreamPartitions

tonytanger:
feat!: rename ListChangeStreamPartitions to GenerateInitialChangeStreamPartitions

Change-Id: Id306df92de00e172ae900a9c4bf95de856edd90f


proto-google-cloud-bigtable-v2/src/main/java/com/google/bigtable/v2/ReadChangeStreamRequestOrBuilder.java
```diff
@@ -110,7 +110,7 @@
    *
    * @return The partition.
    */
-  StreamPartition getPartition();
+  com.google.bigtable.v2.StreamPartition getPartition();
   /**
```
tengzhonger, line 114: Curious: why are these types fully qualified now?

tonytanger: I'm not sure.. But this is how it should be.

PR #1355, opened 2022-08-12
feat: Change CDC related APIs to return ByteStringRange instead of Ro…

tengzhonger:
…wRange

1. Make GenerateInitialChangeStreamPartitions return ByteStringRange
2. Make ChangeStreamContinuationToken::GetRowRange() return ByteStringRange

Also add serializeByteStringRange() & deserializeByteStringRange() methods for ByteStringRange so that they can be used by the CDC beam connector to write to bigtable.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Range.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Range.java
@@ -15,10 +15,13 @@
  */
 package com.google.cloud.bigtable.data.v2.models;
 
+import com.google.api.core.InternalApi;
 import com.google.api.core.InternalExtensionOnly;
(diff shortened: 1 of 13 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Range.java
```diff
@@ -395,6 +398,22 @@ private void writeObject(ObjectOutputStream output) throws IOException {
       output.defaultWriteObject();
     }
 
+    @InternalApi("Used in Changestream beam pipeline.")
+    static ByteString serializeByteStringRange(ByteStringRange byteStringRange) {
```
mutianf, line 402: if this is used in beam it needs to be public

mutianf (approved):

PR #1359, opened 2022-08-15
feat: Return MutationType and bigtable.common.Status instead of raw p…

tengzhonger:
…rotos

This avoid exposing raw protos through CDC classes.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/CloseStream.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/CloseStream.java
@@ -23,6 +23,7 @@ import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
 import com.google.rpc.Status;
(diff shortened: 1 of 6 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/CloseStream.java
```diff
@@ -50,8 +50,8 @@ private CloseStream(Status status, List<StreamContinuationToken> continuationTok
   }
 
   @InternalApi("Used in Changestream beam pipeline.")
-  public Status getStatus() {
-    return this.status;
+  public com.google.cloud.bigtable.common.Status getStatus() {
```
mutianf, line 53: why long import?

tengzhonger: Because I already import `import com.google.rpc.Status;`. CloseStream holds the proto type status (i.e. `com.google.rpc.Status`), and `getStatus()` return the wrapper class(i.e. `com.google.cloud.bigtable.common.Status`)

PR #1398, opened 2022-09-15
feat: copy backup API support

TracyCuiCan:


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientTests.java
```diff
@@ -683,6 +693,78 @@ public void testListBackups() {
...
+  public void testCopyBackup() {
+    // Setup
+    Mockito.when(mockStub.copyBackupOperationCallable())
+        .thenReturn(mockCopyBackupOperationCallable);
+
+    Timestamp startTime = Timestamp.newBuilder().setSeconds(1234).build();
+    Timestamp endTime = Timestamp.newBuilder().setSeconds(5678).build();
+
+    // Use existing adminClient as destination project:
+    String dstProjectId = PROJECT_ID;
```
kolea2, line 706: nit - I would just use `PROJECT_ID` and `INSTANCE_ID` inline and remove these local variables.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
```diff
@@ -303,6 +304,63 @@ public void crossInstanceRestoreTest()
     }
   }
 
+  @Test
+  public void crossInstanceCopyBackupTest()
```
kolea2, line 308: I would add an additional copy backup test that is in the same instance.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClient.java
```diff
@@ -1248,6 +1249,66 @@ public ApiFuture<Void> awaitOptimizeRestoredTableAsync(
...
+   *             .setExpireTime(expireTime);
+   * Backup response = client.copyBackup(request);
+   * }</pre>
+   */
+  public Backup copyBackup(CopyBackupRequest request) {
+    return ApiExceptions.callAndTranslateApiException(copyBackupAsync(request));
+  }
+
+  /**
+   * Creates a copy of backup from an existing backup in a Cloud Bigtable cluster with the specified
```
kolea2, line 1272:
```suggestion
   * Creates a copy of a backup from an existing backup in a Cloud Bigtable cluster with the specified
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,143 @@
...
+      com.google.bigtable.admin.v2.CopyBackupRequest.newBuilder();
+  private final String sourceBackupId;
+  private final String sourceClusterId;
+  private final String sourceInstanceId;
+  private final String sourceProjectId;
+
+  private String destClusterId;
+
+  /**
+   * Create a {@link CopyBackupRequest} object. It assumes the source backup locates in the same
```
kolea2, line 40:
```suggestion
   * Create a {@link CopyBackupRequest} object. It assumes the source backup is located in the same
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,143 @@
...
+   * #of(String, String, String, String) of} method
+   */
+  public static CopyBackupRequest of(String instanceId, String clusterId, String backupId) {
+    CopyBackupRequest request = new CopyBackupRequest(null, instanceId, clusterId, backupId);
+    return request;
+  }
+
+  /**
+   * Create a {@link CopyBackupRequest} object. The source backup could locate in the same or
+   * different culster and/or instance and/or project.
```
kolea2, line 62:
```suggestion
   * different cluster and/or instance and/or project.
```

kolea2 (changes requested):

google-cloud-bigtable-stats/dependency-reduced-pom.xml
```diff
@@ -0,0 +1,408 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
TracyCuiCan: I see that the same change was committed 12 days ago to https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable-stats/pom.xml. This was a valid change back then and should be no change now.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,143 @@
...
+    CopyBackupRequest request = new CopyBackupRequest(null, null, clusterId, backupId);
+    return request;
+  }
+
+  /**
+   * Create a {@link CopyBackupRequest} object. It assumes the source backup locates in the same
+   * project as the destination backup. To copy from a backup in another instance, use {@link
+   * #of(String, String, String, String) of} method
+   */
+  public static CopyBackupRequest of(String instanceId, String clusterId, String backupId) {
```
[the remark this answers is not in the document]
TracyCuiCan: thanks, done!

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
TracyCuiCan: Done.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientTests.java
TracyCuiCan: Done.

google-cloud-bigtable-stats/dependency-reduced-pom.xml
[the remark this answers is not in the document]
kolea2: this is the `dependency-reduced-pom.xml` which gets generated (as opposed to `pom.xml`) and it should be reverted.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,143 @@
...
+  private String destClusterId;
+
+  /**
+   * Create a {@link CopyBackupRequest} object. It assumes the source backup is located in the same
+   * instance as the destination backup. To copy from a backup in another instance, use {@link
+   * #of(String, String, String) of} method. To copy from a backup in another instance, use {@link
+   * #of(String, String, String, String) of} method
+   */
+  public static CopyBackupRequest of(String clusterId, String backupId) {
+    CopyBackupRequest request = new CopyBackupRequest(null, null, clusterId, backupId);
```
kolea2, line 46: This doesn't seem to match up - `clusterId` is the 3rd parameter, but the 3rd parameter should be `instanceId` per the method signature: https://github.com/googleapis/java-bigtable/pull/1398/files#diff-f22bea44852cc477c88df0651b15890d220ccbfdf7f998d0e97bc6c30d409b8bR65

kolea2 (changes requested):

google-cloud-bigtable-stats/dependency-reduced-pom.xml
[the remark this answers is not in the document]
kolea2: (separately we should figure out how to exclude this - maybe in `.gitignore`

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
```diff
@@ -303,6 +304,105 @@ public void crossInstanceRestoreTest()
...
+    try (BigtableTableAdminClient destTableAdmin =
+        testEnvRule.env().getTableAdminClientForInstance(destInstance)) {
+
+      try {
+        CopyBackupRequest req =
+            CopyBackupRequest.of(targetCluster, backupId, testEnvRule.env().getInstanceId())
+                .setBackupId(copiedBackupId)
+                .setClusterId(destCluster)
+                .setExpireTime(expireTime);
+        Backup result = destTableAdmin.copyBackup(req);
```
kolea2, line 381: Can we do some assertions for a working/correct case?

kolea2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
TracyCuiCan: The private method is defined as: 
private CopyBackupRequest(
      @Nullable String sourceProjectId,
      @Nullable String sourceInstanceId,
      @Nonnull String sourceClusterId,
      @Nonnull String sourceBackupId);
The 3rd parameter is ClusterId, the order of parameters of the private method and the public ones are different, should I match them?

kolea2: got it - in this case, the private method should follow the order of the public ones to avoid confusion (like I just had :)). It will also change the order so the `null` parameters will go toward the end, which is more typical and expected.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableBackupIT.java
TracyCuiCan: I believe the assertions are for working/correct cases, the error message is just for when the assertion fails.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
TracyCuiCan: sg, I'll udpate!

google-cloud-bigtable-stats/dependency-reduced-pom.xml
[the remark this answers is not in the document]
TracyCuiCan: I see, I'll remove this file from the PR. Will figure out how to exclude this and make changes in a different PR, sounds good?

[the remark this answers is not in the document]
kolea2: perfect, thanks!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,143 @@
...
+    this.sourceProjectId = sourceProjectId;
+  }
+
+  public CopyBackupRequest setBackupId(String backupId) {
+    Preconditions.checkNotNull(backupId);
+    requestBuilder.setBackupId(backupId);
+    return this;
+  }
+
+  public CopyBackupRequest setClusterId(String clusterId) {
```
kolea2, line 89: maybe rename this to `setDestinationClusterId` and same with the parameter to match the variable and to be clearer to the user?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,143 @@
...
+
+  private String destClusterId;
+
+  /**
+   * Create a {@link CopyBackupRequest} object. It assumes the source backup is located in the same
+   * instance as the destination backup. To copy from a backup in another instance, use {@link
+   * #of(String, String, String) of} method. To copy from a backup in another project, use {@link
+   * #of(String, String, String, String) of} method
+   */
+  public static CopyBackupRequest of(String clusterId, String backupId) {
```
kolea2, line 45: same suggestion here, rename variables to `sourceClusterId` for clarity

kolea2 (approved):
lgtm with just some last suggestions on method and variable naming. Thanks for putting this together!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,144 @@
...
+, null);
+    return request;
+  }
+
+  /**
+   * Create a {@link CopyBackupRequest} object. The source backup could be located in the same or
+   * different cluster and/or instance and/or project.
+   */
+  public static CopyBackupRequest of(
+      String sourceClusterId, String sourceBackupId, String sourceInstanceId, String sourceProjectId) {
```
igorbernstein2, line 66: I think the method names need to be a bit more descriptive...its pretty easy to mess this up and hard to know what field mean what. Also the ordering seems a bit backwards.

Proposal:
// keep of as the bare minimum fields
CopyBackupRequest from(String sourceClusterId, String sourceBackupId)

// for a different instance
CopyBackupRequest fromAnotherInstance(String instanceId, String sourceClusterId, String sourceBackupId)

// for a different project/instance, keeping the params in hierarchical order 
CopyBackupRequest fromAnotherProject(String projectId, String instanceId, String sourceClusterId, String sourceBackupId)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,144 @@
...
+import com.google.cloud.bigtable.admin.v2.internal.NameUtil;
+import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
+import com.google.protobuf.util.Timestamps;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import org.threeten.bp.Instant;
+
+/** Fluent wrapper for {@link com.google.bigtable.admin.v2.CopyBackupRequest} */
```
igorbernstein2, line 28: Please add more documentation here explaining what this does. Currently protobufs are considered an internal implementation detail, so we shouldnt direct users to them. The javadoc here should be able to stand on its own

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,144 @@
...
+    Preconditions.checkNotNull(backupId);
+    requestBuilder.setBackupId(backupId);
+    return this;
+  }
+
+  public CopyBackupRequest setDestinationClusterId(String clusterId) {
+    Preconditions.checkNotNull(clusterId);
+    this.destClusterId = clusterId;
+    return this;
+  }
```
igorbernstein2, line 94: can dest backupid be set without a dest cluster?
If not, it might be better to have a setDestination(clsuterId, backupId) instead

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
```diff
@@ -0,0 +1,137 @@
...
+    requestBuilder.setBackupId(backupId);
+    this.destClusterId = clusterId;
+    return this;
+  }
+
+//  public CopyBackupRequest setDestinationClusterId(String clusterId) {
+//    Preconditions.checkNotNull(clusterId);
+//    this.destClusterId = clusterId;
+//    return this;
+//  }
```
igorbernstein2, line 87: please remove

igorbernstein2 (approved):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
TracyCuiCan: Done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
TracyCuiCan: Added example in the method doc.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
TracyCuiCan: changed this to setDestination(cluster, backup).

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CopyBackupRequest.java
TracyCuiCan: Done

PR #1472, opened 2022-10-21
samples: add delete samples

kolea2:


samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteCheckAndMutateExample.java
```diff
@@ -0,0 +1,44 @@
...
+package com.example.bigtable.deletes;
+
+// [START bigtable_delete_check_and_mutate]
+import com.google.cloud.bigtable.data.v2.BigtableDataClient;
+import com.google.cloud.bigtable.data.v2.models.ConditionalRowMutation;
+import com.google.cloud.bigtable.data.v2.models.Filters;
+import com.google.cloud.bigtable.data.v2.models.Mutation;
+import java.io.IOException;
+
+public class DeleteCheckAndMutateExample {
```
billyjacobson, line 26: I think ConditionalDelete or DeleteConditional might be clearer here, it's just a little wordy

samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteCheckAndMutateExample.java
```diff
@@ -0,0 +1,44 @@
...
+      String tableId,
+      String rowKey,
+      String family,
+      String qualifier) {
+    try (BigtableDataClient dataClient = BigtableDataClient.create(projectId, instanceId)) {
+      Filters.Filter condition = Filters.FILTERS.qualifier().exactMatch(qualifier);
+      Mutation mutation = Mutation.create().deleteCells(family, qualifier);
+      dataClient.checkAndMutateRow(
+          ConditionalRowMutation.create(tableId, rowKey).condition(condition).then(mutation));
+    } catch (IOException e) {
```
billyjacobson, line 39: aren't all the example functions just supposed to throw errors now unless they're super specific? 

https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md#exception-handling

samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteStreamingAndBatchingExample.java
```diff
@@ -0,0 +1,49 @@
...
+import com.google.api.gax.batching.Batcher;
+import com.google.api.gax.rpc.ServerStream;
+import com.google.cloud.bigtable.data.v2.BigtableDataClient;
+import com.google.cloud.bigtable.data.v2.models.Filters;
+import com.google.cloud.bigtable.data.v2.models.Query;
+import com.google.cloud.bigtable.data.v2.models.Row;
+import com.google.cloud.bigtable.data.v2.models.RowMutationEntry;
+import java.io.IOException;
+
+public class DeleteStreamingAndBatchingExample {
```
billyjacobson, line 29: what is the streaming part? could also just be called BatchDelete

samples/snippets/src/test/java/com/example/bigtable/deletes/DeletesTest.java
```diff
@@ -0,0 +1,165 @@
...
+    row = bigtableDataClient.readRow(TABLE_ID, rowKey);
+    Truth.assertThat(row).isNull();
+  }
+
+  @Test
+  public void test3_testStreamingAndBatching() {
+    String rowKey = "phone#4c410523#20190502";
+    Row row = bigtableDataClient.readRow(TABLE_ID, rowKey);
+
+    String familyName = "cell_plan";
```
billyjacobson, line 90: perhaps making these a const at the top could be easier. Also they might be accessible from the base test? I don't remember

billyjacobson (changes requested):
Thank you for these! Let me know if you have any questions

samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteCheckAndMutateExample.java
kolea2: added the exception to the method signature instead.

samples/snippets/src/test/java/com/example/bigtable/deletes/DeletesTest.java
kolea2: replaced some variables with the static ones in the base test. Some, like 'rowKey' are only used in two places at once, so I left them as is.

samples/snippets/src/main/java/com/example/bigtable/deletes/BatchDeleteExample.java
```diff
@@ -20,29 +20,24 @@
...
 import com.google.cloud.bigtable.data.v2.models.RowMutationEntry;
 import java.io.IOException;
 
-public class DeleteStreamingAndBatchingExample {
-  public void streamingAndBatching(
-      String projectId, String instanceId, String tableId, String familyName, String qualifier) {
+public class BatchDeleteExample {
+  public void batchDelete(String projectId, String instanceId, String tableId)
+      throws InterruptedException, IOException {
     try (BigtableDataClient dataClient = BigtableDataClient.create(projectId, instanceId)) {
```
billyjacobson, line 31: I don't think you need the try's for these since there is no catch, then no need for all the extra indentation etc.

samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteDropRowRangeExample.java
```diff
@@ -21,12 +21,10 @@
 import java.io.IOException;
 
 public class DeleteDropRowRangeExample {
```
billyjacobson, line 23: should this be DropRowRangeExample or DeleteRowRangeExample? DeleteDrop seems confusing

samples/snippets/src/test/java/com/example/bigtable/deletes/DeletesTest.java
billyjacobson: Looks good!

samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteColumnFamilyExample.java
```diff
@@ -0,0 +1,34 @@
...
+// [START bigtable_delete_column_family]
+import com.google.cloud.bigtable.data.v2.BigtableDataClient;
+import com.google.cloud.bigtable.data.v2.models.RowMutation;
+import java.io.IOException;
+
+public class DeleteColumnFamilyExample {
+  public void deleteColumnFamily(
+      String projectId, String instanceId, String tableId, String rowKey, String familyName) {
+    try (BigtableDataClient dataClient = BigtableDataClient.create(projectId, instanceId)) {
+      dataClient.mutateRow(RowMutation.create(tableId, rowKey).deleteFamily(familyName));
```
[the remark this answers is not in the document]
billyjacobson: This is probably clarified by the documentation/reading the code – I think I might've been thrown off by it since I didn't know you can delete all the column family data for a row and defaulting to thinking about deleting CF for the table

samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteCheckAndMutateExample.java
```diff
@@ -0,0 +1,44 @@
...
+public class DeleteCheckAndMutateExample {
+  public void checkAndMutate(
+      String projectId,
+      String instanceId,
+      String tableId,
+      String rowKey,
+      String family,
+      String qualifier) {
+    try (BigtableDataClient dataClient = BigtableDataClient.create(projectId, instanceId)) {
+      Filters.Filter condition = Filters.FILTERS.qualifier().exactMatch(qualifier);
```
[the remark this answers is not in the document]
billyjacobson: I assume this is one of the examples you would need to change for the other languages? I do think it would be better to change it since filters with Bigtable can be so confusing, so I'd prefer to show them in a way that makes more sense/adds reason for why you would need it.

samples/snippets/src/main/java/com/example/bigtable/deletes/BatchDeleteExample.java
kolea2: I was following: https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md#client-initialization. Should I undo this?

samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteDropRowRangeExample.java
kolea2: done

samples/snippets/src/main/java/com/example/bigtable/deletes/DeleteCheckAndMutateExample.java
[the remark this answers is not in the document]
kolea2: done!

billyjacobson (approved):

PR #1477, opened 2022-10-25
fix: fix attempt status tag for metrics

mutianf:
We need to fix: https://github.com/googleapis/gax-java/blob/11c4a7caf31de256f29678de829f11f9558adcfd/gax/src/main/java/com/google/api/gax/rpc/ServerStreamingAttemptCallable.java#L379


igorbernstein2 (approved):
:(

PR #1489, opened 2022-11-01
feat: table deletion protection

kolea2:


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/Table.java
```diff
@@ -152,6 +158,10 @@ public List<ColumnFamily> getColumnFamilies() {
     return columnFamilies;
   }
 
+  public boolean isProtected() {
```
TracyCuiCan, line 161: nit: isDeletionProtected

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientTests.java
```diff
@@ -204,6 +212,54 @@ public void testCreateTable() {
     assertThat(result).isEqualTo(Table.fromProto(expectedResponse));
   }
 
+  @Test
+  public void testCreateTableWithDeletionProtection() {
```
TracyCuiCan, line 216: This test doesn't seem to create a table with deletion protection, or maybe I'm missing something?

kolea2: yes great call, thanks :D

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/Table.java
kolea2: This was to be consistent with go - https://github.com/googleapis/google-cloud-go/blob/main/bigtable/bttest/inmem.go#L1330. LMK your thoughts

PR #1498, opened 2022-11-07
feat: add test proxy to java-bigtable client

mutianf:

### DIFF
--- a/test-proxy/README.md
+++ b/test-proxy/README.md
@@ -34,7 +34,7 @@ mvn exec:java -Dexec.mainClass=com.google.cloud.bigtable.testproxy.CbtTestProxyM
 Build and start the proxy with an older version of the client
 
 ```
-mvn clean install -Dbigtable.client.version=2.1.0 -Denforcer.skip
+mvn clean install -Dbigtable.client.version=<client_version> -Denforcer.skip
 mvn exec:java -Dexec.mainClass=com.google.cloud.bigtable.testproxy.CbtTestProxyMain
 ```
 
--- a/test-proxy/README.md
+++ b/test-proxy/README.md
@@ -1,0 +1,46 @@
+# CBT Java Test Proxy
+
+The CBT test proxy is intended for running confromance tests for Cloug Bigtable Java Client.
+
+## Set up 
+
+If you have not already done so, [install golang](https://go.dev/doc/install), then clone the go test library:
+
+```
+git clone https://github.com/googleapis/cloud-bigtable-clients-test.git
+```
+
+## Start test proxy
+
+Build the proxy with the latest version of the client
+
+```
+cd java-bigtable/test-proxy
+mvn clean install
+```
+
+Start the proxy on default port 9999
+
+```
+mvn exec:java -Dexec.mainClass=com.google.cloud.bigtable.testproxy.CbtTestProxyMain
+```
+
+Start the proxy on a different port
+
+```
+mvn exec:java -Dexec.mainClass=com.google.cloud.bigtable.testproxy.CbtTestProxyMain -Dport=1
+```
+
+Build and start the proxy with an older version of the client
+
+```
+mvn clean install -Dbigtable.client.version=2.1.0 -Denforcer.skip
+mvn exec:java -Dexec.mainClass=com.google.cloud.bigtable.testproxy.CbtTestProxyMain
+```
+
+## Run the test cases
+
+```
+cd cloud-bigtable-clients-test/tests
+go test -v -proxy_addr=:9999
+```
--- a/test-proxy/pom.xml
+++ b/test-proxy/pom.xml
@@ -12,11 +12,11 @@
   <parent>
     <artifactId>google-cloud-bigtable-parent</artifactId>
     <groupId>com.google.cloud</groupId>
-    <version>2.15.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
+    <version>2.16.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
   </parent>
 
   <properties>
-    <bigtable.client.version>2.15.1-SNAPSHOT</bigtable.client.version><!-- {x-version-update:google-cloud-bigtable:current} -->
+    <bigtable.client.version>2.16.1-SNAPSHOT</bigtable.client.version><!-- {x-version-update:google-cloud-bigtable:current} -->
   </properties>
 
   <dependencyManagement>
@@ -88,7 +88,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-deploy-plugin</artifactId>
-        <version>3.0.0-M2</version>
         <configuration>
           <skip>true</skip>
         </configuration>
--- a/test-proxy/pom.xml
+++ b/test-proxy/pom.xml
@@ -1,0 +1,126 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.cloud</groupId>
+  <artifactId>google-cloud-bigtable-test-proxy</artifactId>
+  <version>1.0</version>
+  <packaging>jar</packaging>
+  <name>Google Cloud Bigtable Test Proxy</name>
+  <url>https://github.com/googleapis/java-bigtable</url>
+
+  <parent>
+    <artifactId>google-cloud-bigtable-parent</artifactId>
+    <groupId>com.google.cloud</groupId>
+    <version>2.15.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
+  </parent>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-bigtable-bom</artifactId>
+        <version>2.15.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>google-cloud-bigtable-deps-bom</artifactId>
+        <version>2.15.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.google.auto.value</groupId>
+        <artifactId>auto-value</artifactId>
+        <version>1.9</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-bigtable</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-netty</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.grpc</groupId>
+      <artifactId>grpc-stub</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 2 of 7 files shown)


test-proxy/pom.xml
```diff
@@ -0,0 +1,126 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.cloud</groupId>
+  <artifactId>google-cloud-bigtable-test-proxy</artifactId>
+  <version>1.0</version>
+  <packaging>jar</packaging>
+  <name>Google Cloud Bigtable Test Proxy</name>
+  <url>https://github.com/googleapis/java-bigtable</url>
+
```
igorbernstein2, line 10: This definitely needs a description

test-proxy/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+        <artifactId>google-cloud-bigtable-deps-bom</artifactId>
+        <version>2.15.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+      <dependency>
+        <groupId>com.google.auto.value</groupId>
+        <artifactId>auto-value</artifactId>
+        <version>1.9</version>
+      </dependency>
```
igorbernstein2, line 37: Here and everywhere else autovalue is mentioned in the pom:

I think you want to remove this and just add a ENableAutoValue file:

https://github.com/googleapis/java-shared-config/blob/main/pom.xml#L734-L779

test-proxy/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+        <configuration>
+          <protocArtifact>com.google.protobuf:protoc:3.9.0:exe:${os.detected.classifier}</protocArtifact>
+          <pluginId>grpc-java</pluginId>
+          <pluginArtifact>io.grpc:protoc-gen-grpc-java:1.24.0:exe:${os.detected.classifier}</pluginArtifact>
+        </configuration>
+        <executions>
+          <execution>
+            <goals>
+              <goal>compile</goal>
+              <goal>compile-custom</goal>
```
igorbernstein2, line 86: whats compile-custom?

test-proxy/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-shade-plugin</artifactId>
+        <version>3.2.4</version>
+        <executions>
+          <execution>
+            <phase>package</phase>
```
igorbernstein2, line 97: why does this need to be shaded?

test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java
```diff
@@ -0,0 +1,768 @@
...
+      responseObserver.onError(
+          Status.ALREADY_EXISTS
+              .withDescription("Client " + request.getClientId() + " already exists.")
+              .asException());
+      return;
+    }
+
+    BigtableDataSettings.Builder settingsBuilder = BigtableDataSettings.newBuilder();
+    if (request.hasPerOperationTimeout()) {
+      com.google.protobuf.Duration timeoutFromReq = request.getPerOperationTimeout();
```
igorbernstein2, line 225: Can you simplify this to Duration newTimeout = Duration.fromMillis(Durations.toMillis(request.getPerOperationTimeout());

test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java
```diff
@@ -0,0 +1,768 @@
...
+    }
+
+    responseObserver.onNext(CreateClientResponse.getDefaultInstance());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void closeClient(
+      CloseClientRequest request, StreamObserver<CloseClientResponse> responseObserver) {
+    CbtClient client = idClientMap.get(request.getClientId());
```
igorbernstein2, line 260: Why not just call getClient()?

test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java
```diff
@@ -0,0 +1,768 @@
...
+  public void mutateRow(
+      MutateRowRequest request, StreamObserver<MutateRowResult> responseObserver) {
+    CbtClient client = idClientMap.get(request.getClientId());
+    if (client == null) {
+      responseObserver.onError(
+          Status.NOT_FOUND
+              .withDescription("Client " + request.getClientId() + " not found.")
+              .asException());
+      return;
+    }
```
igorbernstein2, line 301: here and everywhere else...just why not just use getCLient?

test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java
```diff
@@ -0,0 +1,768 @@
...
+  }
+
+  private CredentialsProvider getCredentialsProvider() throws IOException {
+    if (credential == null) {
+      return NoCredentialsProvider.create();
+    }
+
+    final ServiceAccountJwtAccessCredentials creds =
+        ServiceAccountJwtAccessCredentials.fromStream(
+            new ByteArrayInputStream(credential.getBytes(UTF_8)));
```
igorbernstein2, line 754: why are you hardcoding jwt creds? I think GoogleCredentials.fromStream() is more appropriate

test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxyMain.java
```diff
@@ -0,0 +1,49 @@
...
+
+  private static final Logger logger = Logger.getLogger(CbtTestProxyMain.class.getName());
+
+  public static void main(String[] args) throws InterruptedException, IOException {
+    int port = Integer.parseInt(System.getProperty("port", "9999"));
+    if (port <= 0) {
+      throw new IllegalArgumentException(String.format("Port %d is not > 0.", port));
+    }
+
+    CbtTestProxy cbtTestProxy = CbtTestProxy.createUnencrypted();
```
igorbernstein2, line 35: should this be in an else?

test-proxy/pom.xml
kolea2: @igorbernstein2 @mutianf are we planning on publishing this module?

mutianf: I think so, is there any concern with adding this in public repo?

test-proxy/pom.xml
mutianf: It's from the example in grpc-java: https://github.com/grpc/grpc-java/blob/v1.50.2/examples/pom.xml#L106

test-proxy/pom.xml
mutianf: To specify the main class to start the proxy I think?

test-proxy/pom.xml
```diff
@@ -0,0 +1,126 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.cloud</groupId>
+  <artifactId>google-cloud-bigtable-test-proxy</artifactId>
+  <version>1.0</version>
```
kolea2, line 6: I would recommend changing this to 0.0.1-SNAPSHOT, and this also needs to be added to versions.txt

kolea2 (changes requested):

test-proxy/pom.xml
mutianf: I don't see samples in versions.txt though?

test-proxy/pom.xml
igorbernstein2: I dont see any reason not to

test-proxy/pom.xml
igorbernstein2: I see compile is for protobufs and custom is for grpc stubs

test-proxy/pom.xml
igorbernstein2: main entry point in the jar can be specified to make it executable:
https://maven.apache.org/shared/maven-archiver/examples/classpath.html#Make

Shading will produce a single jar thats executable. There is no readme here so I'm not sure how this is going to be executed. So I'll list a couple of approaches:

1. Having a bash script that will mvn exec:java the proxy and then start the test suite
2. Have a junit wrapper that starts the proxy at the beginning of the test run and then exec's the test suite

In both cases, we don't need to bundle the dependencies. In fact, bundling the deps prevents us from running the proxy with older clients

test-proxy/pom.xml
```diff
@@ -3,10 +3,11 @@
...
   <artifactId>google-cloud-bigtable-test-proxy</artifactId>
-  <version>1.0</version>
+  <version>0.0.1-SNAPSHOT</version>
   <packaging>jar</packaging>
   <name>Google Cloud Bigtable Test Proxy</name>
   <url>https://github.com/googleapis/java-bigtable</url>
+  <description>Cloud Bigtable Java Client test proxy for running conformance tests.</description>
 
   <parent>
     <artifactId>google-cloud-bigtable-parent</artifactId>
```
liujiongxin, line 13: Does the version tag below accept custom values (i.e., we want to test a version released a while ago)?

liujiongxin (commented):
Overall look good to me! Thanks for the effort!

test-proxy/pom.xml
mutianf: To maven central? I told Kristen that we're not going to 😅

mutianf: Chatted offline, decided to not publishing it for now.

liujiongxin (commented):
A readme file will be very helpful.

Example: https://github.com/googleapis/google-cloud-go/tree/main/bigtable/internal/testproxy, https://github.com/dbolduc/google-cloud-cpp/tree/cbt-test-proxy-dev-flattened/google/cloud/bigtable/cbt_test_proxy

test-proxy/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+              <goal>compile-custom</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <!-- start skip publishing to maven central -->
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <version>3.0.0-M2</version>
```
mutianf, line 91: we probably don't need this version

test-proxy/pom.xml
```diff
@@ -0,0 +1,141 @@
...
+  <version>0.0.1-SNAPSHOT</version>
+  <packaging>jar</packaging>
+  <name>Google Cloud Bigtable Test Proxy</name>
+  <url>https://github.com/googleapis/java-bigtable</url>
+  <description>Cloud Bigtable Java Client test proxy for running conformance tests.</description>
+
+  <parent>
+    <artifactId>google-cloud-bigtable-parent</artifactId>
+    <groupId>com.google.cloud</groupId>
+    <version>2.15.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-bigtable:current} -->
```
mutianf, line 15: update the parent version since we cut a new release

igorbernstein2 (approved):

test-proxy/README.md
```diff
@@ -0,0 +1,46 @@
...
+Start the proxy on a different port
+
+```
+mvn exec:java -Dexec.mainClass=com.google.cloud.bigtable.testproxy.CbtTestProxyMain -Dport=1
+```
+
+Build and start the proxy with an older version of the client
+
+```
+mvn clean install -Dbigtable.client.version=2.1.0 -Denforcer.skip
```
kolea2, line 37: nit - I would replace 2.1.0 with <client_version> so the developer is intentional about which version they pick

kolea2 (approved):
lgtm with one small comment

PR #1503, opened 2022-11-07
chore: turn off fail-fast behavior to prevent cancellation of jobs

mpeddada1:
As seen in https://github.com/googleapis/java-bigtable/pull/1497, unit tests (11) gets cancelled if unit tests (7) fails. This PR prevents workflows from getting cancelled at the first failed job.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,6 +13,7 @@ jobs:
           - 7
           - 8
           - 11
+      fail-fast: false
     steps:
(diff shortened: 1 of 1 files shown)


suztomo (approved):

mpeddada1:
Running into a flaky test with a `Instance and cluster metadata write requests' and limit 'Instance and cluster metadata write requests per day' of service` message. Will try again in a few.

prash-mi (approved):

PR #1529, opened 2022-11-21
test(fix): update the test proxy to set app profile id and connect to emulator correctly

liujiongxin:
This change makes the test proxy work with the hermetic testing that exercises app_profile_id and emulator (https://github.com/googleapis/cloud-bigtable-clients-test)

### DIFF
--- a/test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java
+++ b/test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java
@@ -215,7 +215,11 @@ public class CbtTestProxy extends CloudBigtableV2TestProxyImplBase implements Cl
       return;
     }
 
-    BigtableDataSettings.Builder settingsBuilder = BigtableDataSettings.newBuilder();
+    BigtableDataSettings.Builder settingsBuilder = BigtableDataSettings.newBuilder()
+        .setProjectId(request.getProjectId())
+        .setInstanceId(request.getInstanceId())
+        .setAppProfileId(request.getAppProfileId());
+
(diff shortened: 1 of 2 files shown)


test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java
```diff
@@ -215,7 +215,11 @@ public synchronized void createClient(
       return;
     }
 
-    BigtableDataSettings.Builder settingsBuilder = BigtableDataSettings.newBuilder();
+    BigtableDataSettings.Builder settingsBuilder = BigtableDataSettings.newBuilder()
+        .setProjectId(request.getProjectId())
+        .setInstanceId(request.getInstanceId())
+        .setAppProfileId(request.getAppProfileId());
```
mutianf, line 221: what if the request doesn't have an app profile id?

mutianf (approved):

mutianf:
Please update the commit message and the name of the pr to be:

fix: Update the test proxy to accommodate custom app profile and emulator #1529

kolea2:
> Please update the commit message and the name of the pr to be:
> 
> fix: Update the test proxy to accommodate custom app profile and emulator #1529

Can we prefix these as `test(fix)` to skip them in the release notes?

test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java
liujiongxin: I also add the prefix "test(fix)" to the title and commit.

PR #1546, opened 2022-12-12
test(fix): update the test proxy's pom file to build a shaded jar, also correct the README about building an old version

liujiongxin:
…so correct the README about building an old version

### DIFF
--- a/test-proxy/pom.xml
+++ b/test-proxy/pom.xml
@@ -134,6 +134,38 @@
             <skip>true</skip>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-shade-plugin</artifactId>
+          <executions>
+            <execution>
+              <phase>package</phase>
+              <goals>
+                <goal>shade</goal>
+              </goals>
+              <configuration>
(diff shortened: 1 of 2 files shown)


test-proxy/pom.xml
```diff
@@ -134,6 +134,39 @@
             <skip>true</skip>
           </configuration>
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-shade-plugin</artifactId>
+          <version>3.2.4</version>
```
mutianf, line 140: Do we need the version here? I think it's defined in the bigtable-parent. Can you try build it without the version?

mutianf: > I also updated the README to remove the "build with old release"

But I think we should document it somewhere, and readme is a good place for this? 

> For the other plugins, there are warnings though (they are not introduced by this PR, we may deal with them in the future)

I think you can ignore those.

mutianf (approved):

test-proxy/pom.xml
liujiongxin: Done. Thanks! I will squash the commits after it looks good.

PR #1604, opened 2023-01-25
feat: track the latency a request is queued on the grpc channel

mutianf:
For all operations, it tracks the latency of the request getting queued on a grpc channel. For batch operations, it's the aggregated value of grpc channel queued latency and batcher flow control latencies.


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableGrpcStreamTracer.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package com.google.cloud.bigtable.data.v2.stub.metrics;
+
+import com.google.common.base.Stopwatch;
+import io.grpc.Attributes;
+import io.grpc.ClientStreamTracer;
+import io.grpc.Metadata;
+import java.util.concurrent.TimeUnit;
+
+/** Records the time a request is blocked on the grpc channel */
```
igorbernstein2, line 24: I woudnt say blocked, but enqueued in a grpc channel queue

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/Util.java
```diff
@@ -197,4 +199,18 @@ static void recordMetricsFromMetadata(
     // Record gfe metrics
     tracer.recordGfeMetadata(latency, throwable);
   }
+
+  static GrpcCallContext addHandlerToCallContext(
+      ApiCallContext context, GrpcResponseMetadata responseMetadata, BigtableTracer tracer) {
+    if (context instanceof GrpcCallContext) {
+      // context should always be an instance of GrpcCallContext. Sanity check just in case.
```
igorbernstein2, line 206: If this should never happen, I think we should throw an exception not swallow the error so that we can fix our assumptions

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -484,7 +484,7 @@ private <RowT> UnaryCallable<Query, List<RowT>> createBulkReadRowsCallable(
         new TracedBatcherUnaryCallable<>(readRowsUserCallable.all());
 
     UnaryCallable<Query, List<RowT>> withBigtableTracer =
-        new BigtableTracerUnaryCallable<>(tracedBatcher);
+        new BigtableTracerUnaryCallable<>(tracedBatcher, true);
```
igorbernstein2, line 487: using a boolean here is hard to read...consider using 2 different factory methods:
BigtableTracerUnaryCallable.createForBatchedCallable

igorbernstein2: Better yet, make a separate BigtableTracerBatchedCallable that inherits from the Unary one

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/Util.java
```diff
@@ -197,4 +199,18 @@ static void recordMetricsFromMetadata(
     // Record gfe metrics
     tracer.recordGfeMetadata(latency, throwable);
   }
+
+  static GrpcCallContext addHandlerToCallContext(
```
igorbernstein2, line 203: Method name is too generic. You can only know what handler is when you read the method body. The method name should should directed to the caller as the audience. Maybe injectBigtableStreamTracer?

igorbernstein2: Also ad javadoc that explains that this bridges gRPC's stream tracing to bigtable tracing

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableGrpcStreamTracer.java
igorbernstein2: I would also use the docs to describe that this a bridge between gRPC stream tracing and Bigtable tracing. It primary purpose is to measure the transition time between asking gRPC to start an RPC and gRPC actually serializing that RPC

google-cloud-bigtable-stats/src/test/java/com/google/cloud/bigtable/stats/StatsRecorderWrapperTest.java
```diff
@@ -92,6 +92,7 @@ public void testStreamingOperation() throws InterruptedException {
     recorderWrapper.putGfeMissingHeaders(connectivityErrorCount);
     recorderWrapper.putFirstResponseLatencies(firstResponseLatency);
     recorderWrapper.putBatchRequestThrottled(throttlingLatency);
+    recorderWrapper.putRequestBlockedOnChannel(throttlingLatency);
```
igorbernstein2, line 95: I would rename this to putGrpcQueueDuration

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableTracerStreamingCallable.java
```diff
@@ -60,7 +67,15 @@ public void call(
...
               responseObserver, (BigtableTracer) context.getTracer(), responseMetadata);
-      innerCallable.call(request, innerObserver, responseMetadata.addHandlers(context));
+      if (batching) {
+        innerCallable.call(request, innerObserver, responseMetadata.addHandlers(context));
+      } else {
+        innerCallable.call(
+            request,
+            innerObserver,
+            Util.addHandlerToCallContext(
+                context, responseMetadata, (BigtableTracer) context.getTracer()));
```
igorbernstein2, line 77: The difference here is too subtle.
1. extract the difference out:
if (batching) {
  ctx = Util.addHandlerToCallContext(
                context, responseMetadata, (BigtableTracer) context.getTracer()))
} else {
  ctx = responseMetadata.addHandlers(context)
}

innerCallable.call(request, observer, ctx).


2. Add comments to the if statement as to why it needs different handling

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableGrpcStreamTracer.java
mutianf: Hmmm which doc are you referring to?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracerTest.java
```diff
@@ -180,6 +213,29 @@ public void sendHeaders(Metadata headers) {
...
+          @Override
+          @SuppressWarnings("rawtypes")
+          public ManagedChannelBuilder apply(ManagedChannelBuilder builder) {
+            if (oldConfigurator != null) {
+              builder = oldConfigurator.apply(builder);
+            }
+            return builder.intercept(clientInterceptor);
+          }
+        };
+    channelProvider.setChannelConfigurator(newConfigurator);
```
igorbernstein2, line 236: I think a lambda would be easier to read?

```java
channelProvider.setChannelConfigurator((builder) -> {
  if (oldConfigurator != null) {
     builder = oldConfigurator.apply(builder);
   }
return builder.intercept(clientInterceptor)
});

```

igorbernstein2 (approved):
lgtm after nit

igorbernstein2 (approved):

PR #1655, opened 2023-02-28
Add getNewPartitions method to CloseStream for Bigtable ChangeStream

tengzhonger:


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/CloseStream.java
```diff
@@ -46,6 +49,13 @@ public static CloseStream fromProto(@Nonnull ReadChangeStreamResponse.CloseStrea
         closeStream.getStatus(),
         closeStream.getContinuationTokensList().stream()
             .map(ChangeStreamContinuationToken::fromProto)
+            .collect(ImmutableList.toImmutableList()),
+        closeStream.getNewPartitionsList().stream()
```
mutianf, line 53: do we need to check if status is OK then new partition should be empty?  what will happen in this case?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
```diff
@@ -122,6 +122,15 @@ private StreamContinuationToken createStreamContinuationToken(@Nonnull String to
         .build();
   }
 
+  private StreamPartition createNewPartitionForCloseStream() {
```
mutianf, line 125: Is there a retry test that retries only the second partition? should it only retry the second partition?

tengzhonger: We don't need to worry about this because we don't retry after CloseStream: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamResumptionStrategy.java#L58

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/CloseStream.java
tengzhonger: The API should guarantee that only one of the following can happen:
1) Status==OK, new_partitions==empty, continuation_tokens==empty
2) Status!=OK, new_partitions!=empty, continuation_tokens!=empty, in this case new_partitions.size() == continuation_tokens.size()

Added some Preconditions.checkState

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
mutianf: But the comment also says: "will return an OK status right after sending a CloseStream.", but sounds like with this new change, server doesn't always return OK?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/CloseStream.java
```diff
@@ -35,8 +37,22 @@ public abstract class CloseStream implements ChangeStreamRecord, Serializable {
 
   private static CloseStream create(
       com.google.rpc.Status status,
-      List<ChangeStreamContinuationToken> changeStreamContinuationTokens) {
-    return new AutoValue_CloseStream(Status.fromProto(status), changeStreamContinuationTokens);
+      List<ChangeStreamContinuationToken> changeStreamContinuationTokens,
+      List<ByteStringRange> newPartitions) {
+    if (status.getCode() == 0) {
```
mutianf, line 42: I think you can simplify this

Preconditions.checkState((status.getCode() == 0 && changeStreamContinuationTokens.isEmpty) || (status.getCode() != 0 && !changeStreamContinuationToken.isEmpty())

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/changestream/ReadChangeStreamRetryTest.java
tengzhonger: Oh sorry for the confusion. The CloseStream.Status is different from the final status the server returns. If a CloseStream is returned, it's either one of the following:
1) CloseStream.Status==OK, which means the read finishes successfully, server will return an OK right after the CloseStream.
2) CloseStream.Status!=OK, CloseStream.NewPartitions/ContinuationTokens != empty, which means the client(i.e. beam) should start a new ReadChangeStream call with the new partitions. Server will still return an OK right after.

So in either case, server will return an OK right after CloseStream, and there's no need to retry after receiving a CloseStream.

I've updated the comment as well.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/CloseStream.java
tengzhonger: Discussed offline. Let's keep two checks because I want to use different error messages.

mutianf (approved):

PR #1698, opened 2023-04-07
test(fix): add a dummy test to the CI pipeline which doesn't block PR…

liujiongxin:
… merge. It will be completed to run client-binding behavioral test.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,3 +87,24 @@ jobs:
     - run: .kokoro/build.sh
       env:
         JOB_TYPE: clirr
+  behavioral:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/checkout@v3
+      with:
+        repository: googleapis/cloud-bigtable-clients-test
+        ref: main
+        path: behavioral-test
+    - uses: actions/setup-java@v3
+      with:
+        distribution: zulu
+        java-version: 11
+    - uses: actions/setup-go@v3
+      with:
+        go-version: '>=1.20.2'
+    - run: java -version
+    - run: go version
+    - run: .kokoro/build.sh
+      env:
+        JOB_TYPE: behavioral
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -112,6 +112,13 @@ clirr)
     mvn -B -ntp -Denforcer.skip=true clirr:check
     RETURN_CODE=$?
     ;;
+behavioral)
+    # TODO: complete the testing script
+    # Temporary operation: locate the test folder and test proxy folder
+    ls ${scriptDir}/behavioral-test
+    ls test-proxy
+    RETURN_CODE=$?
+    ;;
 *)
     ;;
 esac


.github/workflows/ci.yaml
```diff
@@ -87,3 +87,24 @@ jobs:
...
+    - uses: actions/checkout@v3
+    - uses: actions/checkout@v3
+      with:
+        repository: googleapis/cloud-bigtable-clients-test
+        ref: main
+        path: cloud-bigtable-clients-test
+    - uses: actions/setup-java@v3
+      with:
+        distribution: zulu
+        java-version: 11
```
kolea2, line 102: This should use 8 to match what we use in `maven-compile-plugin` https://github.com/googleapis/java-shared-config/blob/main/pom.xml#L169-L170

.kokoro/build.sh
```diff
@@ -112,6 +112,14 @@ clirr)
     mvn -B -ntp -Denforcer.skip=true clirr:check
     RETURN_CODE=$?
     ;;
+behavioral)
+    # TODO: complete the testing script
```
kolea2, line 116: it's not clear to me why this TODO can't be done in this PR?

kolea2 (changes requested):

.kokoro/build.sh
liujiongxin: I was a little conservative, I can complete it in one PR.

kolea2 (approved):
LGTM, please update PR title and update owlbot.py so this doesn't get overwritten: https://github.com/googleapis/java-bigtable/blob/main/owlbot.py#L97-L112 also would like a review from @mutianf

.kokoro/build.sh
```diff
@@ -112,6 +112,25 @@ clirr)
     mvn -B -ntp -Denforcer.skip=true clirr:check
     RETURN_CODE=$?
     ;;
+conformance)
+    # Build and start the proxy in a separate process
+    pushd .
+    cd test-proxy
+    mvn clean install
```
mutianf, line 119: I think we can skip the other tests here: 
```suggestion
    mvn clean install -DskipTests
```

.kokoro/build.sh
```diff
@@ -112,6 +112,25 @@ clirr)
...
+    nohup java -Dport=9999 -jar target/google-cloud-bigtable-test-proxy-0.0.1-SNAPSHOT.jar &
+    proxyPID=$!
+    popd
+
+    # Invoke the test
+    pushd .
+    cd cloud-bigtable-clients-test/tests
```
mutianf, line 126: Where is this `cloud-bigtable-clients-test`  directory?  I don't see it under test-proxy?

liujiongxin: It's checked out in the ci.yaml:
- uses: actions/checkout@v3
      with:
        repository: googleapis/cloud-bigtable-clients-test
        ref: main
        path: cloud-bigtable-clients-test

The end result is like this:

java-bigtable |
                        |--- test_proxy
                        |--- cloud-bigtable-clients-test

When the if branch is being executed, the current dir "." is at the root (i.e. java-bigtable/)

.kokoro/build.sh
liujiongxin: Done

mutianf (approved):

liujiongxin:
ci.yaml and build.sh are auto-generated by https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/java_library

I need to create separate yaml and script files for conformance test.

PR #1706, opened 2023-04-12
chore: remove enforce-declared-grpc-and-proto-version enforcer check

kolea2:


igorbernstein2 (changes requested):
Now that we arent trying to keep a uniform version, can you rename the properties?

protoc.version -> test-protoc.version
grpc.version -> test-protoc-grpc.version
protobuf.version -> test-protoc-protobuf.version

google-cloud-bigtable/pom.xml
```diff
@@ -39,9 +39,9 @@
 
     <!-- These are needed to compile the protobuf used by Changestream merging acceptance test. We have enforcer rule
      below to make sure that the versions here match the ones we pull in via the shared bom. -->
```
igorbernstein2, line 41: Please update the comment as well :)

it looks like test-protobuf.version is not used, maybe inline its value in test-protoc.version and remove it

kolea2: done

igorbernstein2 (approved):

PR #1708, opened 2023-04-13
feat: add experimental support for reverse scans public preview

igorbernstein2:
This adds a reversed boolean to Query, which will allow endusers to stream rows in reverse order.

Example:
```java
Query query = Query.create("alphabet").range("a", "z").limit(3);
ServerStream<Row> results = client.readRows(query);

for (Row row : results) {
  System.out.println(row.getKey().toStringUtf8());
}
// Prints z, y, x
```
This its currently blocked on generated protos which will be autogenerated soon

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigtable/Reads.java
+++ b/samples/snippets/src/main/java/com/example/bigtable/Reads.java
@@ -24,6 +24,7 @@ import com.google.api.gax.rpc.ServerStream;
 import com.google.cloud.bigtable.data.v2.BigtableDataClient;
 import com.google.cloud.bigtable.data.v2.models.Filters;
 import com.google.cloud.bigtable.data.v2.models.Query;
+import com.google.cloud.bigtable.data.v2.models.Range.ByteStringRange;
 import com.google.cloud.bigtable.data.v2.models.Row;
 import com.google.cloud.bigtable.data.v2.models.RowCell;
 import java.io.IOException;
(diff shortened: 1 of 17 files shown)


samples/snippets/src/main/java/com/example/bigtable/Reads.java
```diff
@@ -200,6 +201,39 @@ public static void readPrefix(String projectId, String instanceId, String tableI
...
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
+    // the "close" method on the client to safely clean up any remaining background resources.
+    try (BigtableDataClient dataClient = BigtableDataClient.create(projectId, instanceId)) {
+      Query query = Query.create(tableId)
+          .reversed(true)
+          .limit(2)
+          .prefix("phone#4c410523")
+          .range(ByteStringRange.unbounded()
+              .startClosed("phone#5c10102")
```
mutianf, line 223: startClosed and end closed is the same key, maybe change to different keys? So customers can know if start key should be < end key or > end key because it's reversed?

mutianf: Or maybe in the javadoc example in Query.java update it to:

```
   * <pre>{@code
   * query
   *   .range(ByteStringRange.unbounded().startKeyClosed("a").endOpen("c"))
   *   .limit(10)
   *   .reversed(true)
   * }</pre>
```

mutianf (approved):

igorbernstein2:
closing in favor of #7891 and will address feedback there

samples/snippets/src/main/java/com/example/bigtable/Reads.java
igorbernstein2: I updated the sample to use different keys

igorbernstein2: (in https://github.com/googleapis/java-bigtable/pull/1711)

PR #1719, opened 2023-04-20
feat: print debug headers when channel priming fails

mutianf:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
@@ -15,18 +15,27 @@
  */
 package com.google.cloud.bigtable.data.v2.stub;
 
+import static io.grpc.Metadata.Key;
+
 import com.google.api.core.BetaApi;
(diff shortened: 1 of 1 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
```diff
@@ -106,6 +132,11 @@ private void sendPrimeRequests(ManagedChannel managedChannel) throws IOException
           e = e.getCause();
         }
         LOG.warning(String.format("Failed to prime channel: %s", e));
+        if (responseMetadata.getMetadata() != null) {
+          LOG.warning(
+              "Encrypted debug header is "
+                  + responseMetadata.getMetadata().get(ENCRYPTED_HEADERS_KEY));
```
igorbernstein2, line 138: I think this needs to be more user friendly. "Channel priming failed, which implies a connection issue. When opening a support case, include this encrypted debug values in the support case"

igorbernstein2 (approved):

igorbernstein2 (changes requested):
I think we need to rethink this approach. If the connection fails, it most liekly will not have gfe headers. :(

PR #1762, opened 2023-05-25
docs(samples): add bigtable filter snippet

Sita04:
1. Add a filter snippet to read row using a filter
2. Add region tags to two existing snippets

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigtable/Filters.java
+++ b/samples/snippets/src/main/java/com/example/bigtable/Filters.java
@@ -366,7 +366,8 @@ public class Filters {
     // once, and can be reused for multiple requests. After completing all of your requests, call
     // the "close" method on the client to safely clean up any remaining background resources.
     try (BigtableDataClient dataClient = BigtableDataClient.create(projectId, instanceId)) {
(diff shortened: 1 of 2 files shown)


samples/snippets/src/main/java/com/example/bigtable/Filters.java
```diff
@@ -354,6 +359,26 @@ public static void filterComposingCondition(String projectId, String instanceId,
   // [END bigtable_filters_composing_condition]
   // [END_EXCLUDE]
 
+  // [START bigtable_hw_get_with_filter]
+  private static void readRowFilter(
+      String projectId, String instanceId, String tableId, Filter filter) {
+    // Initialize client that will be used to send requests. This client only needs to be created
+    // once, and can be reused for multiple requests. After completing all of your requests, call
```
Shabirmean, line 366: I think we can remove the second line here since try with resources will automatically, call close at the end of the block!

Shabirmean (approved):

samples/snippets/src/main/java/com/example/bigtable/Filters.java
Sita04: Sure. Thanks

PR #1768, opened 2023-05-25
docs(samples): remove client initialization

Sita04:
Remove multiple client initializations as the snippets are not used standalone

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigtable/HelloWorld.java
+++ b/samples/snippets/src/main/java/com/example/bigtable/HelloWorld.java
@@ -59,8 +59,6 @@ public class HelloWorld {
   private static final String COLUMN_QUALIFIER_GREETING = "greeting";
   private static final String COLUMN_QUALIFIER_NAME = "name";
   private static final String ROW_KEY_PREFIX = "rowKey";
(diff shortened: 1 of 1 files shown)


samples/snippets/src/main/java/com/example/bigtable/HelloWorld.java
```diff
@@ -221,48 +217,32 @@ public List<Row> readTable() {
   }
 
   // [START bigtable_hw_create_filter]
-  public static void filterLimitCellsPerCol(String projectId, String instanceId, String tableId) {
+  public void filterLimitCellsPerCol(String tableId) {
```
Shabirmean, line 220: Should we make the following methods private as well?
- readTable
- readSpecificCells
- readSingleRow
- writeToTable
- close

Shabirmean (approved):

samples/snippets/src/main/java/com/example/bigtable/HelloWorld.java
Sita04: No, the public methods are being called by tests as well.
So, would prefer to keep it as such.

PR #1788, opened 2023-06-09
feat: migrate to OTEL exporter

mutianf:
Changes in this PR are independent from Bigtable. It only interacts with cloud monitoring API. So there's no rollback plan. The correctness of the metrics conversion from OTEL to cloud monitoring format are tested in `BigtableCloudMonitoringExporterTest` unit test. The end to end IT test will be added to the following PR.


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -0,0 +1,234 @@
...
+        return DOUBLE;
+      case HISTOGRAM:
+      case EXPONENTIAL_HISTOGRAM:
+        return DISTRIBUTION;
+      default:
+        return ValueType.UNRECOGNIZED;
+    }
+  }
+
+  private static Timestamp convertTimestamp(long epochNanos) {
```
igorbernstein2, line 186: I think protobuf already has a utility that does is `Timestamps.fromNanos`

[fromNanos](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/util/Timestamps.html#fromNanos-long-)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -0,0 +1,234 @@
...
+
+    return builder.build();
+  }
+
+  static String getProjectId(PointData pointData) {
+    return pointData.getAttributes().get(PROJECT_ID);
+  }
+
+  private static MetricKind convertMetricKind(MetricData metricData) {
+    switch (metricData.getType()) {
```
igorbernstein2, line 137: It seems like all of our supported metrics CUMULATIVE AggregationTemporality. Why not make it the first check in this function?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -0,0 +1,234 @@
...
+    // If not the expected format then generate a random number.
+    if (jvmName.indexOf('@') < 1) {
+      String hostname = "localhost";
+      try {
+        hostname = InetAddress.getLocalHost().getHostName();
+      } catch (UnknownHostException e) {
+        logger.log(Level.INFO, "Unable to get the hostname.", e);
+      }
+      // Generate a random number and use the same format "random_number@hostname".
+      return "java-" + new SecureRandom().nextInt() + "@" + hostname;
```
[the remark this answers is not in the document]
mutianf: Yeah, we added a UUID when the jvm name is pid@hostname: https://github.com/googleapis/java-bigtable/commit/3bfc7bc246488a37b4470e3a75ffdd34f6761804

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  private static MetricKind convertMetricKind(MetricData metricData) {
+    switch (metricData.getType()) {
+      case HISTOGRAM:
+      case EXPONENTIAL_HISTOGRAM:
+        return convertHistogramDataType(metricData.getHistogramData());
+      case LONG_GAUGE:
+      case DOUBLE_GAUGE:
+        return GAUGE;
```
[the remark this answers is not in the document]
mutianf: We don't currently .. but want to add it here in case we add a GAUGE metric later..

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
mutianf: AggregationTemporality is not an option for all data types, so I need to first get the data type and then check the AggregationTemporality :(

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,125 @@
...
+  BigtableCloudMonitoringExporter(
+      MetricServiceClient client, MonitoredResource monitoredResource, String taskId) {
+    this.client = client;
+    this.monitoredResource = monitoredResource;
+    this.taskId = taskId;
+  }
+
+  @Override
+  public CompletableResultCode export(Collection<MetricData> collection) {
+    Map<String, List<TimeSeries>> projectToTimeSeries;
```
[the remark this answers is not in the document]
mutianf: There's a problem with openTelemetry when I create metrics I can't have special charactors (I think "/") in the metric name .. I'll need to add prefix to the metrics. And also, since everything else doesn't have MetricDescriptor created in monarch, I don't think unrelevant metrics can be exported anyways

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,131 @@
...
+              .collect(Collectors.toList());
+
+      ProjectName projectName = ProjectName.of(projectId);
+      CreateTimeSeriesRequest request =
+          CreateTimeSeriesRequest.newBuilder()
+              .setName(projectName.toString())
+              .addAllTimeSeries(timeSeries)
+              .build();
+
+      this.client.createServiceTimeSeriesCallable().futureCall(request);
```
igorbernstein2, line 109: shouldn't this be mapped to a CompletableResultCode? ie instantiate the CompletableResultCode and call succeed/fail when the future finishes?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    return CompletableResultCode.ofSuccess();
+  }
+
+  @Override
+  public CompletableResultCode flush() {
+    return CompletableResultCode.ofSuccess();
```
igorbernstein2, line 117: This is a little dangerous, this should probably return a CompletableResultCode that completes when all of the outstanding RPCs finish

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  @Override
+  public CompletableResultCode shutdown() {
```
igorbernstein2, line 121: does this need to flush? I would imagine that you'd want to: set a closed flag so that new export calls are rejected, then flush, then return flush's CompletableResultCode

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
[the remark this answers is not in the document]
igorbernstein2: why dont we use uuid always?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,131 @@
...
+
+  private static final String RESOURCE_TYPE = "bigtable_client_raw";
+
+  static BigtableCloudMonitoringExporter create(String projectId, Credentials credentials)
+      throws IOException {
+    MetricServiceSettings.Builder settingsBuilder = MetricServiceSettings.newBuilder();
+    settingsBuilder.setCredentialsProvider(FixedCredentialsProvider.create(credentials));
+
+    org.threeten.bp.Duration timeout = Duration.ofMinutes(1);
+    settingsBuilder.createServiceTimeSeriesSettings().setSimpleTimeoutNoRetries(timeout);
```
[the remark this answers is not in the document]
mutianf: So looks like createServiceTimeseries and createTimeseries are not retryable. So I wonder if they're not idempotent. createTimeSeries didn't have a timeout, so i'm setting the default timeout to be 1 minute here.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,172 @@
...
+    this.client = client;
+    this.monitoredResource = monitoredResource;
+    this.taskId = taskId;
+    this.projectId = projectId;
+  }
+
+  @Override
+  public CompletableResultCode export(Collection<MetricData> collection) {
+    if (isShutdown.get()) {
+      return CompletableResultCode.ofFailure();
```
igorbernstein2, line 94: might be good to print a log warning

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,172 @@
...
+  private final MetricServiceClient client;
+
+  private final String projectId;
+  private final String taskId;
+  private final MonitoredResource monitoredResource;
+  private AtomicBoolean isShutdown = new AtomicBoolean(false);
+
+  private static final String RESOURCE_TYPE = "bigtable_client_raw";
+
+  private CompletableResultCode lastCode;
```
igorbernstein2, line 61: Can you rename this to be a more descriptive? lastExportResultCode? Also it seems like this will be accessed from different threads, so needs some synchronization

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,172 @@
...
+
+    lastCode = new CompletableResultCode();
+
+    List<TimeSeries> allTimeSeries = new ArrayList<>();
+    for (MetricData metricData : collection) {
+      if (!metricData.getInstrumentationScopeInfo().getName().equals("bigtable.googleapis.com")) {
+        continue;
+      }
+
+      CompletableResultCode code = new CompletableResultCode();
```
igorbernstein2, line 111: is this used?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,172 @@
...
+      return CompletableResultCode.ofFailure();
+    }
+    if (!collection.stream()
+        .flatMap(metricData -> metricData.getData().getPoints().stream())
+        .allMatch(pd -> BigtableExporterUtils.getProjectId(pd).equals(projectId))) {
+      logger.log(Level.WARNING, "Metric data has different a projectId. Skip exporting.");
+      return CompletableResultCode.ofFailure();
+    }
+
+    lastCode = new CompletableResultCode();
```
igorbernstein2, line 103: please move this closer to where its used

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,172 @@
...
+
+    List<TimeSeries> allTimeSeries = new ArrayList<>();
+    for (MetricData metricData : collection) {
```
igorbernstein2, line 106: Why not move this loop to BIgtableExporterUtils? ie the split of converting a point to timeseries vs converting a list of them seems arbitrary

igorbernstein2 (changes requested):
Since this interacts with the real world, I think you need an integration test

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -0,0 +1,255 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+/** Utils to convert OpenTelemetry types to Cloud Monitoring API types. */
+class BigtableExporterUtils {
+
+  private static final Logger logger = Logger.getLogger(BigtableExporterUtils.class.getName());
+
+  static String getDefaultTaskValue() {
```
igorbernstein2, line 73: Please update the comment its currently a lie. I would promote it to javadoc:

In most cases this should look like `java-${UUID}@${hostname}`. The hostname will be retrieved from the jvm name but fallback on the local hostname.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -0,0 +1,255 @@
...
+      } catch (UnknownHostException e) {
+        logger.log(Level.INFO, "Unable to get the hostname.", e);
+      }
+      // Generate a random number and use the same format "random_number@hostname".
+      return "java-" + UUID.randomUUID() + "@" + hostname;
+    }
+    return "java-" + UUID.randomUUID() + jvmName;
+  }
+
+  private static final Set<AttributeKey<String>> PROMOTED_RESOURCE_LABELS =
```
igorbernstein2, line 90: please keep constants at the top of the class

igorbernstein2: Also add a comment explaining that these are metric labels that will be promoted to the bigtable_table monitored resource

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -0,0 +1,255 @@
...
+        continue;
+      }
+
+      List<TimeSeries> timeSeries =
+          metricData.getData().getPoints().stream()
+              .map(
+                  pointData ->
+                      convertPointToTimeSeries(metricData, pointData, taskId, monitoredResource))
+              .collect(Collectors.toList());
+      allTimeSeries.addAll(timeSeries);
```
igorbernstein2, line 108: you can slightly simplify to:

```
metricData.getData().getPoints().stream()
  .map(...)
  .forEach(allTimeSeries::add)
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -0,0 +1,255 @@
...
+          attributeKey.getKey(), String.valueOf(attributes.get(attributeKey)));
+    }
+
+    // Populate the rest of the metric labels
+    attributes.forEach(
+        (key, value) -> {
+          if (!PROMOTED_RESOURCE_LABELS.contains(key) && value != null) {
+            metricLabels.put(key.getKey(), String.valueOf(value));
+          }
+        });
```
igorbernstein2, line 135: these 2 are basically doing the same thing, but using 2 different loop styles. The only practical difference where bucket to put the entry into.

I think it would be clearer to do this:

```
for(String key : attributes) {
  String value = String.valueOf(attributes.get(key));
  if (PROMOTED_RESOURCE_LABELS.contains(key)) {
    resourceBuilder.putLabel(key, value);
  } else {
    metricLabels.put(...)
}
```

igorbernstein2: One more thing: why bother with intermedia metricLabels Map? why not just add it directly to the Metric?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,157 @@
...
+
+  @Override
+  public CompletableResultCode export(Collection<MetricData> collection) {
+    if (isShutdown.get()) {
+      logger.log(Level.WARNING, "Exporter is shutting down");
+      return CompletableResultCode.ofFailure();
+    }
+    if (!collection.stream()
+        .flatMap(metricData -> metricData.getData().getPoints().stream())
+        .allMatch(pd -> BigtableExporterUtils.getProjectId(pd).equals(projectId))) {
```
igorbernstein2, line 97: this might throw an NPE. Might be better to reverse this to projectId.equals(...)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsAttributes.java
```diff
@@ -0,0 +1,34 @@
...
+
+import io.opentelemetry.api.common.AttributeKey;
+
+class BuiltinMetricsAttributes {
+
+  public static final AttributeKey<String> PROJECT_ID = AttributeKey.stringKey("project_id");
+  public static final AttributeKey<String> INSTANCE_ID = AttributeKey.stringKey("instance");
+  public static final AttributeKey<String> TABLE_ID = AttributeKey.stringKey("table");
+  public static final AttributeKey<String> CLUSTER_ID = AttributeKey.stringKey("cluster");
+  public static final AttributeKey<String> ZONE_ID = AttributeKey.stringKey("zone");
```
igorbernstein2, line 26: why public in a pkg protected class?

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,125 @@
...
+import io.opentelemetry.sdk.metrics.export.MetricExporter;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import org.threeten.bp.Duration;
+
+final class BigtableCloudMonitoringExporter implements MetricExporter {
```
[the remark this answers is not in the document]
igorbernstein2: Please add a bit more info:

/** 
Bigtable Cloud Monitoring OpenTelemetry Exporter. 

The exporter will look for all bigtable owned metrics under the InstrumentationScopeInfo bigtable.googleapis.com and upload it via the Google Cloud Monitoring api.
*/

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,174 @@
...
+      logger.log(Level.WARNING, "shutdown is called multiple times");
+      return CompletableResultCode.ofSuccess();
+    }
+    CompletableResultCode resultCode = flush();
+    try {
+      client.shutdown();
+    } catch (Throwable e) {
+      logger.log(Level.WARNING, "failed to shutdown the monitoring client", e);
+      return CompletableResultCode.ofFailure();
+    }
```
igorbernstein2, line 162: you need to chain these:
```
CompletableResultCode closeFuture = new CompletableResultCode();
CompletableResultCode flushFuture = flush();

flushFuture.whenComplere(() -> {
    Exception closeError = null;
  try {
    client.shutdown();
  } catch(RuntimeException e) {
    closeError = e;
  }
  if (closeError == null) {
    closeFuture.succeed();
  } else {
    log....
   closeFuture.fail()
  } 
})
```

Otherwise you will close the client while its flushing

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -0,0 +1,174 @@
...
+    return resultCode;
+  }
+
+  /**
+   * For cloud monarch always return CUMULATIVE to keep track of the cumulative value of a metric
```
igorbernstein2, line 167: I think you mean Google Cloud Monitoring

igorbernstein2 (approved):

PR #1808, opened 2023-06-26
fix: fix batch mutation limit

mutianf:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
@@ -69,6 +69,9 @@ public final class BulkMutation implements Serializable, Cloneable {
   public BulkMutation add(@Nonnull String rowKey, @Nonnull Mutation mutation) {
     Preconditions.checkNotNull(rowKey);
     Preconditions.checkNotNull(mutation);
+    Preconditions.checkArgument(
+        mutation.getMutations().size() <= 100000,
+        "Too many mutations, got " + mutation.getMutations().size() + ", limit is 100000");
 
     return add(ByteString.copyFromUtf8(rowKey), mutation);
   }
@@ -80,6 +83,9 @@ public final class BulkMutation implements Serializable, Cloneable {
   public BulkMutation add(@Nonnull ByteString rowKey, @Nonnull Mutation mutation) {
     Preconditions.checkNotNull(rowKey);
     Preconditions.checkNotNull(mutation);
+    Preconditions.checkArgument(
+        mutation.getMutations().size() <= 100000,
+        "Too many mutations, got " + mutation.getMutations().size() + ", limit is 100000");
 
     builder.addEntries(
         MutateRowsRequest.Entry.newBuilder()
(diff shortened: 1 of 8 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
```diff
@@ -69,6 +69,9 @@ private void writeObject(ObjectOutputStream output) throws IOException {
   public BulkMutation add(@Nonnull String rowKey, @Nonnull Mutation mutation) {
     Preconditions.checkNotNull(rowKey);
     Preconditions.checkNotNull(mutation);
+    Preconditions.checkArgument(
```
igorbernstein2, line 72: do you want to also check the total mutation count across rows as well?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/mutaterows/MutateRowsBatchResource.java
```diff
@@ -0,0 +1,59 @@
...
+  static MutateRowsBatchResource create(long elementCount, long byteCount, long mutationCount) {
+    return new AutoValue_MutateRowsBatchResource(elementCount, byteCount, mutationCount);
+  }
+
+  @Override
+  public BatchResource add(BatchResource batchResource) {
+    Preconditions.checkArgument(
+        batchResource instanceof MutateRowsBatchResource,
+        "Expected MutateRowsBatchResource, got " + batchResource.getClass());
+    MutateRowsBatchResource mutateRowsResource = (MutateRowsBatchResource) batchResource;
```
igorbernstein2, line 38: nit. please add a newline to make it easier to read

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
mutianf: Hmm I'll need to iterate through all the entries I added to BulkMutation and calculate the sum everytime I add some elements. Is it worth? I think as long as someone is using the batcher this should be handled by the batcher?

igorbernstein2: why not keep a running sum?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
```diff
@@ -81,6 +86,11 @@ public BulkMutation add(@Nonnull ByteString rowKey, @Nonnull Mutation mutation)
     Preconditions.checkNotNull(rowKey);
     Preconditions.checkNotNull(mutation);
 
+    this.mutationCountSum += mutation.getMutations().size();
+    Preconditions.checkArgument(
+        mutationCountSum <= MAX_MUTATION,
+        String.format("Too many mutations, got %s, limit is %s", mutationCountSum, MAX_MUTATION));
```
igorbernstein2, line 92: you need to check the precondition before modifying state

igorbernstein2 (approved):

mutianf:
I removed all the checks in client because of https://google.aip.dev/client-libraries/4223

PR #1823, opened 2023-06-28
feat: change stream retention to create and update table

tonytanger:
Follow up to https://github.com/googleapis/java-bigtable/pull/1821 adding update table support.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CreateTableRequest.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CreateTableRequest.java
@@ -16,6 +16,7 @@
 package com.google.cloud.bigtable.admin.v2.models;
 
 import com.google.api.core.InternalApi;
+import com.google.bigtable.admin.v2.ChangeStreamConfig;
 import com.google.bigtable.admin.v2.ColumnFamily;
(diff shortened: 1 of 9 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CreateTableRequest.java
```diff
@@ -76,6 +78,22 @@ public CreateTableRequest addSplit(ByteString key) {
     return this;
   }
 
+  /** Add change stream retention period between 1 day and 7 days */
+  public CreateTableRequest addChangeStreamRetention(Duration retention) {
+    Preconditions.checkNotNull(retention);
```
mutianf, line 83: should we add precondition to check retention is between 1 and 7?

tonytanger: I wasn't sure what the philosophy is. Checking here would allow us to fail faster but if we widen the retention period then we have to change on the client side as well.

kolea2: Per https://google.aip.dev/client-libraries/4223, I think we should rely on the server side validation.

mutianf (approved):

PR #1858, opened 2023-07-26
deps: shared-dependencies 3.13.0

burkedavison:


kolea2 (approved):

google-cloud-bigtable-deps-bom/pom.xml
```diff
@@ -66,7 +66,7 @@
       <dependency>
         <groupId>com.google.cloud</groupId>
         <artifactId>google-cloud-shared-dependencies</artifactId>
-        <version>3.11.0</version>
+        <version>3.13.0</version>
```
kolea2, line 69: google-cloud-monitoring-bom should also probably be updated to the corresponding version (I see some test failures)

burkedavison: google-cloud-monitoring-bom is at 3.20.0, which corresponds to the current LTS version for this library.

We _shouldn't_ have compatibility issues using shared-deps 3.13.0 with a library that was built using 3.11.0, but I'm investigating.

PR #1934, opened 2023-09-22
chore: add a partial template

JoeWang1127:


JoeWang1127:
Confirm the partial file works.

PR #2026, opened 2023-12-06
feat: handle retry info so client respect the delay server sets

mutianf:
Creates a new retry algorithm that checks if the throwable has a RetryInfo field, and use the retry_delay in the RetryInfo as the backoff time for the next retry.

Test and rollback plan: https://docs.google.com/document/d/1Kim-80vUSDgYxDVrFtkl8NkV-SsW8KAor_Ca1rpvbyo/edit?resourcekey=0-oV2CmU7_sZEMpPMFF9ZSOA&tab=t.0#bookmark=id.mtpgcav79p4u


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ApiResultRetryAlgorithm.java
```diff
@@ -43,11 +72,30 @@ public boolean shouldRetry(
...
+      return extractRetryDelay(previousThrowable) != null
+          || ((previousThrowable instanceof ApiException)
+              && context
+                  .getRetryableCodes()
+                  .contains(((ApiException) previousThrowable).getStatusCode().getCode()));
     }
     return shouldRetry(previousThrowable, previousResponse);
   }
+
+  public static Duration extractRetryDelay(Throwable throwable) {
```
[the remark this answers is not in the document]
igorbernstein2: I see its needed for MutateRows.
So given the current design of gax, if this were to be upstreamed, this would be an isntance method on ApiException. However I understand that you cant fork ApiException. So as a compromise I think it would be better to create a class `ApiExceptions` and add the method there, with a comment saying that this belongs on ApiException.

Also, I dont think you want to expose the actual delay to veneer code, but only if the error is retryable. So maybe something like this:

ApiExceptions {
  // TODO: this should replace the existing ApiException#isRetryable() method, but that cant be done in bigtable, so this lives here for now
  static boolean isRetryable2(ApiException e) {
     if (extractRetryDelay(e) != null) {
        return true;
     }
     return e.isRetryable();
}

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ApiResultRetryAlgorithm.java
```diff
@@ -18,17 +18,46 @@
...
+
+  @Override
+  public TimedAttemptSettings createNextAttempt(
+      Throwable prevThrowable, ResponseT prevResponse, TimedAttemptSettings prevSettings) {
+    Duration retryDelay = extractRetryDelay(prevThrowable);
+    if (retryDelay != null) {
+      return prevSettings
+          .toBuilder()
+          .setRandomizedRetryDelay(retryDelay)
+          .setAttemptCount(prevSettings.getAttemptCount() + 1)
```
igorbernstein2, line 49: I wonder if its possible to align with grpc native retries reset the next non-server instructed delay to the initial?

https://github.com/grpc/proposal/blob/master/A6-client-retries.md#pushback

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ApiResultRetryAlgorithm.java
```diff
@@ -18,17 +18,46 @@
...
+/**
+ * For internal use, public for technical reasons. This retry algorithm checks the metadata of an
+ * exception for additional error details. If the metadata has a RetryInfo field, use the retry
+ * delay to set the wait time between attempts.
+ */
 @InternalApi
 public class ApiResultRetryAlgorithm<ResponseT> extends BasicResultRetryAlgorithm<ResponseT> {
 
+  private static final Metadata.Key<RetryInfo> KEY_RETRY_INFO =
+      ProtoUtils.keyForProto(RetryInfo.getDefaultInstance());
```
igorbernstein2, line 39: What exactly does this do? what will be the actual metadata key?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/RetryInfoTest.java
```diff
@@ -0,0 +1,377 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class RetryInfoTest {
+
+  @Rule public GrpcServerRule serverRule = new GrpcServerRule();
+
+  private static final Metadata.Key<RetryInfo> RETRY_INFO_KEY =
```
igorbernstein2, line 74: make the original package private and reference that so that we can keep them aligned

igorbernstein2 (changes requested):
Also please update the pr description with:
- how this was tested e2e
- a link to a rollback doc

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -893,6 +921,42 @@ public Map<String, String> getJwtAudienceMapping() {
...
+    @BetaApi("Routing cookie is not currently stable and may change in the future")
+    public boolean getEnableRoutingCookie() {
+      return enableRoutingCookie;
+    }
+
+    /**
+     * Gets if RetryInfo is enabled. If true, client bases retry decision and back off time on
+     * server returned RetryInfo value. Otherwise, client uses {@link RetrySettings}.
+     */
+    public boolean getEnableRetryInfo() {
```
igorbernstein2, line 956: BetaApi

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -762,11 +763,18 @@ public Map<String, String> extract(MutateRowsRequest mutateRowsRequest) {
...
+    RetryAlgorithm<Void> retryAlgorithm;
+    ExponentialRetryAlgorithm exponentialRetryAlgorithm =
+        new ExponentialRetryAlgorithm(
+            settings.bulkMutateRowsSettings().getRetrySettings(), clientContext.getClock());
+    if (settings.getEnableRetryInfo()) {
+      retryAlgorithm =
+          new RetryAlgorithm<>(new RetryInfoRetryAlgorithm<>(), exponentialRetryAlgorithm);
+    } else {
+      retryAlgorithm =
+          new RetryAlgorithm<>(new ApiResultRetryAlgorithm<>(), exponentialRetryAlgorithm);
```
igorbernstein2, line 775: Can you pull RetryAlgorithm out and have the branches assign a ResultRetryAlgorithm? and then after the if instantiate the RetryAlgorithm?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -325,6 +327,14 @@ public boolean getEnableRoutingCookie() {
     return enableRoutingCookie;
   }
 
+  /**
+   * Gets if RetryInfo is enabled. If true, client bases retry decision and back off time on server
+   * returned RetryInfo value. Otherwise, client uses {@link RetrySettings}.
+   */
+  public boolean getEnableRetryInfo() {
```
igorbernstein2, line 334: BetaApi

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -918,6 +931,15 @@ public Builder setEnableRoutingCookie(boolean enableRoutingCookie) {
       return this;
     }
 
+    /**
+     * Sets if RetryInfo is enabled. If true, client bases retry decision and back off time on
+     * server returned RetryInfo value. Otherwise, client uses {@link RetrySettings}.
+     */
+    public Builder setEnableRetryInfo(boolean enableRetryInfo) {
```
igorbernstein2, line 938: BetaApi

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ApiException.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.gaxx.retrying;
+
+import com.google.api.core.InternalApi;
+
+// TODO: move this to gax later
+@InternalApi
+public class ApiException {
```
igorbernstein2, line 22: This should be ApiException**s**. Also please add a private ctor

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/RetryInfoRetryAlgorithm.java
```diff
@@ -0,0 +1,110 @@
...
+      // of codes that should be retried.
+      return ((previousThrowable instanceof ApiException)
+          && context
+              .getRetryableCodes()
+              .contains(((ApiException) previousThrowable).getStatusCode().getCode()));
+    }
+    return shouldRetry(previousThrowable, previousResponse);
+  }
+
+  static Duration extractRetryDelay(Throwable throwable) {
```
igorbernstein2, line 93: Please annotate the param & return with Nullable

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/RetryInfoTest.java
```diff
@@ -0,0 +1,668 @@
...
+public class RetryInfoTest {
+
+  @Rule public GrpcServerRule serverRule = new GrpcServerRule();
+
+  private FakeBigtableService service;
+  private BigtableDataClient client;
+  private BigtableDataSettings.Builder settings;
+
+  private AtomicInteger attemptCounter = new AtomicInteger();
+  private Duration delay = Duration.newBuilder().setSeconds(1).setNanos(0).build();
```
igorbernstein2, line 85: nit. use the fully qualified name here and important the java one in the tests...you use it a lot more

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/RetryInfoTest.java
```diff
@@ -0,0 +1,668 @@
...
+      try {
+        client.generateInitialChangeStreamPartitions("table").iterator().hasNext();
+      } catch (ApiException e) {
+        assertThat(e.getStatusCode()).isEqualTo(exception.getStatusCode());
+      }
+      assertThat(attemptCounter.get()).isEqualTo(1);
+    }
+  }
+
+  private void createRetryableExceptionWithDelay(Duration delay) {
```
igorbernstein2, line 517: nit I would name this enqueueRetryableException instead of create

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/RetryInfoTest.java
```diff
@@ -0,0 +1,668 @@
...
+    try (BigtableDataClient client = BigtableDataClient.create(settings.build())) {
+      createRetryableExceptionWithDelay(delay);
+      Stopwatch stopwatch = Stopwatch.createStarted();
+      client.readRow("table", "row");
+      stopwatch.stop();
+
+      assertThat(attemptCounter.get()).isEqualTo(2);
+      assertThat(stopwatch.elapsed()).isLessThan(java.time.Duration.ofSeconds(delay.getSeconds()));
+
+      attemptCounter.set(0);
```
igorbernstein2, line 145: should these be separate tests?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/RetryInfoTest.java
```diff
@@ -0,0 +1,668 @@
...
+      ApiException exception = createNonRetryableExceptionWithDelay(delay);
+      try {
+        client.readRow("table", "row");
+      } catch (ApiException e) {
+        assertThat(e.getStatusCode()).isEqualTo(exception.getStatusCode());
+      }
+      assertThat(attemptCounter.get()).isEqualTo(1);
+    }
+  }
```
igorbernstein2, line 154: there is a lot of repetition for each method here. Unless I'm mistaken can you just write a single test that takes a Runnable and then just pass the method invocation there?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ApiException.java
mutianf: In gax it is ApiException I think https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/ApiException.java

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/RetryInfoTest.java
mutianf: I think these are still testing RetryInfo related behaviors so it's ok to leave them here?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ApiException.java
igorbernstein2: In general the pattern is that the instance class is singular and the helper utility with static methods is plural.
In this case you are adding latter. Eventually when you upstream, you will make it an instance method and it will be in ApiException. But in the transitional state I think plural makes sense

igorbernstein2 (approved):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ApiExceptions.java
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.gaxx.retrying;
+
+import com.google.api.core.InternalApi;
+
+// TODO: move this to gax later
```
kolea2, line 20: do we have a plan for when we're moving this to gax?

PR #2027, opened 2023-12-07
chore: add gax package private classes

mutianf:
This PR is a direct copy of package private classes in https://github.com/googleapis/sdk-platform-java/tree/main/gax-java/gax/src/main/java/com/google/api/gax/rpc to temporarily fork it.

Callables.java only forked retry related methods.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/Callables.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/Callables.java
@@ -1,31 +1,17 @@
 /*
- * Copyright 2017 Google LLC
+ * Copyright 2023 Google LLC
  *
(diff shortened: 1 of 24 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/Callables.java
```diff
@@ -0,0 +1,103 @@
...
+import java.util.Collection;
+
+// TODO: remove this once ApiResultRetryAlgorithm is added to gax.
+/**
+ * Class with utility methods to create callable objects using provided settings.
+ *
+ * <p>The callable objects wrap a given direct callable with features like retry and exception
+ * translation.
+ */
+@BetaApi
```
igorbernstein2, line 39: Please mark this as internal since its temporary

PR #2041, opened 2023-12-28
fix: fix RetryInfo algorithm and tests

mutianf:
Gax already parses ErrorDetails from an error response and add the error details to ApiException. Fix the RetryInfoAlgorithm to handle this correctly and the test to send error responses with the correct format.

Also fixed MutateRowsAttemptCallable to not use RetryInfoAlgorithm with the setting is disabled. 


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/mutaterows/MutateRowsAttemptCallable.java
```diff
@@ -250,7 +258,12 @@ private void handleAttemptError(Throwable rpcError) {
     currentRequest = builder.build();
     originalIndexes = newOriginalIndexes;
 
-    throw new MutateRowsException(rpcError, allFailures.build(), entryError.isRetryable());
+    ErrorDetails errorDetails = null;
+    if (rpcError instanceof ApiException) {
+      errorDetails = ((ApiException) rpcError).getErrorDetails();
+    }
+    throw new MutateRowsException(
+        rpcError, allFailures.build(), entryError.isRetryable(), errorDetails);
```
igorbernstein2, line 266: You are already passing rpcError to MutateRowsException, why do you need extract the error details first and pass them in parallel?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/mutaterows/MutateRowsAttemptCallable.java
```diff
@@ -268,6 +281,8 @@ private void handleAttemptSuccess(List<MutateRowsResponse> responses) {
     List<Integer> newOriginalIndexes = Lists.newArrayList();
     boolean[] seenIndices = new boolean[currentRequest.getEntriesCount()];
 
+    ErrorDetails errorDetails = null;
```
igorbernstein2, line 284: I'm quite uncertain about embedding error details per entry. More so, I'm really weirded out about using the last entry-with-error-details to be the overall error details for the MutateRowsException.

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/mutaterows/MutateRowsAttemptCallable.java
mutianf: MutateRowsException extends ApiException, which now has an ErrorDetails field. RetryInfoRetryAlgorithm checks .getErrorDetails() to determine if it should use retry delay for error handling. Passing the errorDetails here so I don't need to add special handling for MutateRowsException in RetryInfoRetryAlgorithm, which will be moved to gax later.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/mutaterows/MutateRowsAttemptCallable.java
```diff
@@ -230,14 +236,16 @@ private void handleAttemptError(Throwable rpcError) {
     Builder builder = lastRequest.toBuilder().clearEntries();
     List<Integer> newOriginalIndexes = Lists.newArrayList();
 
+    TimedAttemptSettings nextAttemptSettings =
```
igorbernstein2, line 239: why do you need a local var for this? why not jsut assign to the instance var attemptSettings?

igorbernstein2 (approved):
lgtm after nit

PR #2049, opened 2024-01-11
test: update conformance settings

mutianf:
In the conformance test, create the client with retryInfo and routing cookie disabled. Run all the tests except for the tests that specifically test the features, which are expected to fail because client is not handling them.

Also create another client with both features enabled and expected the tests to succeed.

### DIFF
--- a/.kokoro/conformance.sh
+++ b/.kokoro/conformance.sh
@@ -59,8 +59,10 @@ do
     echo "Testing the client with all optional features enabled..."
     configFlag="--enable_features_all"
   else
-    echo "Testing the client with default settings for optional features..."
-    configFlag=""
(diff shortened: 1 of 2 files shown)


.kokoro/conformance.sh
```diff
@@ -59,8 +59,10 @@ do
     echo "Testing the client with all optional features enabled..."
     configFlag="--enable_features_all"
   else
-    echo "Testing the client with default settings for optional features..."
-    configFlag=""
+    echo "Testing the client with default settings for optional features..."\
```
liujiongxin, line 62: can the trailing "\" be removed?

liujiongxin (approved):

PR #2062, opened 2024-01-19
feat: append version to the client name in client-side metrics.

### LINKED ISSUE #2061: Pass client version in client-side metrics
Pass along the client version when building the client-side metrics.

We can do this by appending the client version to the client name in StatsRecorderWrapper when building the context.

rkaregar:

Fixes #2061 ☕️


google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/StatsRecorderWrapper.java
```diff
@@ -59,9 +59,10 @@ public StatsRecorderWrapper(
     this.operationMeasureMap = statsRecorder.newMeasureMap();
   }
 
-  public void recordOperation(String status, String tableId, String zone, String cluster) {
+  public void recordOperation(
+      String status, String tableId, String zone, String cluster, String clientVersion) {
     TagContextBuilder tagCtx =
-        newTagContextBuilder(tableId, zone, cluster)
+        newTagContextBuilder(tableId, zone, cluster, clientVersion)
```
mutianf, line 65: In the BuiltinMetricsTracerFactory, you can add client version to statsAttributes so we don't need to update the APIs. I think this will simplify this PR a lot.

mutianf (approved):

PR #2088, opened 2024-02-01
feat: Define the metrics for collecting per connection error count.

### LINKED ISSUE #2087: Define a builtin client-side metric to count the number of errors per connection per minute
We want to define a new builtin metric to periodically count the number of RPCs with error per connection, for better diagnostics.

rkaregar:

Fixes #2087 ☕️

### DIFF
--- a/google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/BuiltinMeasureConstants.java
+++ b/google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/BuiltinMeasureConstants.java
@@ -29,6 +29,20 @@ class BuiltinMeasureConstants {
   static final TagKey ZONE = TagKey.create("zone");
   static final TagKey CLIENT_UID = TagKey.create("client_uid");
 
+  // Monitored resource TagKeys for gce_instance
+  static final TagKey GCE_PROJECT_ID = TagKey.create("gce_project_id");
+  static final TagKey GCE_INSTANCE_ID = TagKey.create("gce_instance_id");
+  static final TagKey GCE_ZONE = TagKey.create("gce_zone");
+
+  // Monitored resource TagKeys for gke_container
+  static final TagKey GKE_PROJECT_ID = TagKey.create("gke_project_id");
+  static final TagKey GKE_CLUSTER_NAME = TagKey.create("gke_cluster_name");
+  static final TagKey GKE_NAMESPACE_ID = TagKey.create("gke_namespace_id");
+  static final TagKey GKE_INSTANCE_ID = TagKey.create("gke_instance_id");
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/BuiltinMeasureConstants.java
```diff
@@ -29,6 +29,20 @@ class BuiltinMeasureConstants {
   static final TagKey ZONE = TagKey.create("zone");
   static final TagKey CLIENT_UID = TagKey.create("client_uid");
 
+  // Monitored resource TagKeys for gce_instance
+  static final TagKey GCE_PROJECT_ID = TagKey.create("gce_project_id");
```
mutianf, line 33: Do we need to define these keys here? Or this logic could stay in the exporter when we're constructing the detected monitored resource?

google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/BuiltinViewConstants.java
```diff
@@ -183,4 +173,29 @@ class BuiltinViewConstants {
...
           ImmutableList.of(
               PROJECT_ID, INSTANCE_ID, APP_PROFILE, METHOD, CLIENT_NAME, CLUSTER, ZONE, TABLE));
+
+  static final View PER_CONNECTION_ERROR_COUNT_VIEW =
+      View.create(
+          View.Name.create("bigtable.googleapis.com/internal/client/per_connection_error_count"),
+          "Distribution of counts of channels per 'error count per minute'.",
+          PER_CONNECTION_ERROR_COUNT,
+          PER_CONNECTION_ERROR_COUNT_AGGREGATION,
+          ImmutableList.of(
```
mutianf, line 183: Same here, I don't think we need to tag the view with GKE / GCE tag keys. These values are unknown when we populate the metrics.

google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/BuiltinViewConstants.java
```diff
@@ -59,6 +42,13 @@ class BuiltinViewConstants {
                   1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0, 15.0, 20.0, 30.0, 40.0, 50.0,
                   100.0)));
 
+  private static final Aggregation PER_CONNECTION_ERROR_COUNT_AGGREGATION =
```
mutianf, line 45: This only goes up to 30k per connection, would that be enough? I would increase the upper bound

mutianf (approved):

PR #2094, opened 2024-02-02
feat: create the backbone of counting errors per connection each minute.

### LINKED ISSUE #2093: Create infrastructure for counting errors per connection
We want to add a background thread that intercepts RPCs and counts the number of errors per minute per channel. (This will later be exported as a metric.)

rkaregar:

Fixes #2093 ☕️


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -236,6 +233,39 @@ public static ClientContext createClientContext(EnhancedBigtableStubSettings set
     return ClientContext.create(builder.build());
   }
 
+  private static void setInterceptors(
+      InstantiatingGrpcChannelProvider.Builder transportProvider,
+      EnhancedBigtableStubSettings.Builder settings) {
+    Set<ConnectionErrorCountInterceptor> interceptors =
+        setupConnectionErrorCountInterceptors(settings);
+    if (transportProvider != null) {
```
mutianf, line 241: Let's have this method return a list of interceptors.

```
ImmutableList.<ClientInterceptor>Builder builder = ImmutableList.builder()
    .add(new  ConnectionErrorCountInterceptor(interceptors))
if (settings.getEnableRoutingCookie()) {
    builder.add(new CookiesInterceptor());
}

return builder.build();

```

And on line 208 you can do 
```
if (transportProvider instanceof InstantiatingGrpcTransportProvider) {
   ((InstantiatingGrpcTransportProvider) transportProvider).setInterceptor(() -> ...)
}
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/ConnectionErrorCountInterceptor.java
```diff
@@ -0,0 +1,73 @@
...
+import io.grpc.ForwardingClientCall;
+import io.grpc.ForwardingClientCallListener;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/** An interceptor which counts the number of failed responses for a channel. */
+class ConnectionErrorCountInterceptor implements ClientInterceptor {
```
mutianf, line 31: As discussed in the meeting, let's have the interceptor provider creates the set of interceptors and have it accessible by both this class and the background task class.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/CountErrorsPerInterceptorTask.java
```diff
@@ -0,0 +1,48 @@
...
+    this.interceptors = interceptors;
+    this.statsRecorderWrapper = StatsWrapper.createRecorder(null, null, builtinAttributes);
+  }
+
+  @Override
+  public void run() {
+    for (ConnectionErrorCountInterceptor interceptor : interceptors) {
+      int errors = interceptor.getAndResetNumOfErrors();
+      int successes = interceptor.getAndResetNumOfSuccesses();
+      // We avoid keeping track of inactive connections (i.e., without any failed or successful
```
mutianf, line 39: will inactive connections get cleaned up from the set? I think we should clean up inactive connections to avoid the set growing unbounded

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -126,9 +127,9 @@
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
mutianf, line 130: don't do `import *`. You can configure this in your intellij, see https://doc.nuxeo.com/corg/intellij-idea-setup/#:~:text=To%20avoid%20IntelliJ%20IDEA%20replacing,Editor%20%3E%20Code%20Style%20%3E%20Java.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -236,6 +236,38 @@ public static ClientContext createClientContext(EnhancedBigtableStubSettings set
...
+
+    if (settings.getEnableRoutingCookie()) {
+      // TODO: this also need to be added to BigtableClientFactory
+      builder.add(new CookiesInterceptor());
+    }
+
+    return builder.build();
+  }
+
+  private static Set<ConnectionErrorCountInterceptor> setupConnectionErrorCountInterceptors(
```
mutianf, line 257: I think the naming of this method is a bit confusing, this is really set up the background task instead of setting up the interceptors. Let's rename this method to "setupConnectionErrorCountTask"

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -236,6 +236,38 @@ public static ClientContext createClientContext(EnhancedBigtableStubSettings set
...
+      EnhancedBigtableStubSettings.Builder settings) {
+    Set<ConnectionErrorCountInterceptor> interceptors =
+        Collections.newSetFromMap(new WeakHashMap<>());
+    ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
```
mutianf, line 261: @igorbernstein2  do you think we should use the background executor for this?

mutianf: discussed with Igor, we think this should be `ScheduledExecutorService scheduler = settings.getBackgroundExecutorProvider().getExecutor();`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/CountErrorsPerInterceptorTask.java
mutianf: Also chatted with Igor and I don't think we can use WeakHashMap cause it's gonna be accessed from multiple thread. Probably need to use a ConcurrentHashMap here and clean up the stale connections manually :(

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -204,10 +210,13 @@ public static ClientContext createClientContext(EnhancedBigtableStubSettings set
...
-      // TODO: this also need to be added to BigtableClientFactory
-      // patch cookies interceptor
-      transportProvider.setInterceptorProvider(() -> ImmutableList.of(new CookiesInterceptor()));
+    Set<ConnectionErrorCountInterceptor> connectionErrorCountInterceptors =
+        Collections.newSetFromMap(new WeakHashMap<>());
+    setupConnectionErrorCountTask(builder, connectionErrorCountInterceptors);
+
+    if (transportProvider != null) {
+      transportProvider.setInterceptorProvider(
+          () -> getInterceptors(builder, connectionErrorCountInterceptors));
```
mutianf, line 219: Another bug here (carried over from my implementation :(). If someone already set an interceptor on the transport provider, they will lose their settings. We should get the interceptors the user set and merge it.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/CountErrorsPerInterceptorTask.java
rkaregar: looks like there is a Collections.synchronizedSet() method that does it! :)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
rkaregar: discussed offline, the interceptor is a private field in InstantiatingGrpcChannelProvider without a getter. Its getInterceptors() method only gets called inside createSingleChannel(), so seems like we can't access the original one if it's been set. Leaving this as is for now.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/CountErrorsPerInterceptorTask.java
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.data.v2.stub;
+
+import com.google.cloud.bigtable.stats.StatsRecorderWrapper;
+import com.google.cloud.bigtable.stats.StatsWrapper;
+import com.google.common.collect.ImmutableMap;
+import java.util.Set;
+
+class CountErrorsPerInterceptorTask implements Runnable {
```
mutianf, line 23: nit: also add a javadoc for this method. Something like "A background task that goes through all the channels and updates the Errors per connection metric"

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/CountErrorsPerInterceptorTask.java
```diff
@@ -0,0 +1,54 @@
...
+  private final Set<ConnectionErrorCountInterceptor> interceptors;
+  private StatsRecorderWrapper statsRecorderWrapper;
+
+  public void setStatsRecorderWrapper(StatsRecorderWrapper statsRecorderWrapper) {
```
mutianf, line 27: add a `@VisibleForTesting` annotation.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/CountErrorsPerInterceptorTask.java
```diff
@@ -0,0 +1,54 @@
...
+    this.statsRecorderWrapper = statsRecorderWrapper;
+  }
+
+  public CountErrorsPerInterceptorTask(
+      Set<ConnectionErrorCountInterceptor> interceptors,
+      ImmutableMap<String, String> builtinAttributes) {
+    this.interceptors = interceptors;
+    this.statsRecorderWrapper = StatsWrapper.createRecorder(null, null, builtinAttributes);
```
mutianf, line 35: Also add a comment here, something like: "We're only interacting with putAndRecordPerConnectionErrorCount method so OperationType and SpanName won't matter"

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+    ScheduledExecutorService executors = Mockito.mock(ScheduledExecutorService.class);
+    builder =
+        BigtableDataSettings.newBuilderForEmulator(server.getPort())
+            .stubSettings()
+            .setBackgroundExecutorProvider(FixedExecutorProvider.create(executors))
+            .setProjectId("fake-project")
+            .setInstanceId("fake-instance");
+    runnableCaptor = ArgumentCaptor.forClass(Runnable.class);
+    System.out.println("rez1 " + executors + " 2 = " + builder);
```
mutianf, line 63: nit: clean up the debug log :P

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
```diff
@@ -0,0 +1,261 @@
...
+        .putAndRecordPerConnectionErrorCount(errorCountCaptor.capture());
+    runInterceptorTasksAndAssertCount(1);
+    List<Long> allErrorCounts = errorCountCaptor.getAllValues();
+    assertThat(allErrorCounts.size()).isEqualTo(1);
+    assertThat(allErrorCounts.get(0)).isEqualTo(errorCount);
+  }
+
+  @Test
+  public void readWithTwoStubs() throws Exception {
+    EnhancedBigtableStub stub1 = EnhancedBigtableStub.create(builder.build());
```
mutianf, line 109: Instead of creating 2 stubs, you can set the number of channels in EnhancedBigtableStubSettings. For example: `stubSettingsBuilder.setTransportChannelProvider(((InstantiatingGrpcChannelProvider) stubSettingsBuilder.getTransportChannelProvider()).toBuilder().setChannelPoolSettings(ChannelPoolSettings.staticallySized(2)).build());` this gives you 2 channels.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
rkaregar: I had searched for all occurrences of "reza", but forgotten about "rez" 🤣

mutianf (approved):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
igorbernstein2: I think you can work around this by using the ChannelCOnfigurator instead of InterceptorProvider and set the interceptor directly on the ManagedChannelBuilder

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/ConnectionMetricsTracker.java
```diff
@@ -0,0 +1,64 @@
...
+        Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()));
+
+    startConnectionErrorCountTracker();
+  }
+
+  private void startConnectionErrorCountTracker() {
+    ScheduledExecutorService scheduler = builder.getBackgroundExecutorProvider().getExecutor();
+    scheduler.scheduleAtFixedRate(
+        new CountErrorsPerInterceptorTask(
+            connectionErrorCountInterceptors, interceptorsLock, builtinAttributes),
```
mutianf, line 50: passing in the interceptorsLock is a bit weird, can we create a new lock in the other class?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/ConnectionMetricsTracker.java
```diff
@@ -0,0 +1,64 @@
...
+        0,
+        PER_CONNECTION_ERROR_COUNT_PERIOD_SECONDS,
+        TimeUnit.SECONDS);
+  }
+
+  ClientInterceptor getInterceptor() {
+    ConnectionErrorCountInterceptor connectionErrorCountInterceptor =
+        new ConnectionErrorCountInterceptor();
+    synchronized (interceptorsLock) {
+      connectionErrorCountInterceptors.add(connectionErrorCountInterceptor);
```
mutianf, line 60: connectionErrorCountInterceptors is already a synchronized set, so I don't think we need to add a lock around add()?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/CountErrorsPerInterceptorTask.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  CountErrorsPerInterceptorTask(
+      Set<ConnectionErrorCountInterceptor> interceptors,
+      Object interceptorsLock,
+      ImmutableMap<String, String> builtinAttributes) {
+    this.interceptors = interceptors;
+    this.interceptorsLock = interceptorsLock;
+    // We only interact with the putAndRecordPerConnectionErrorCount method, so OperationType and
+    // SpanName won't matter.
```
mutianf, line 46: This comment is not relevant anymore :) we can clean it up.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -204,10 +204,25 @@ public static ClientContext createClientContext(EnhancedBigtableStubSettings set
...
-      transportProvider.setInterceptorProvider(() -> ImmutableList.of(new CookiesInterceptor()));
+    ConnectionMetricsTracker connectionMetricsTracker =
+        new ConnectionMetricsTracker(builder, createBuiltinAttributes(builder));
+    if (transportProvider != null) {
+      ApiFunction<ManagedChannelBuilder, ManagedChannelBuilder> oldChannelConfigurator =
+          transportProvider.getChannelConfigurator();
+      transportProvider.setChannelConfigurator(
+          managedChannelBuilder -> {
+            if (settings.getEnableRoutingCookie()) {
+              // TODO: this also need to be added to BigtableClientFactory
```
mutianf, line 215: this comment can be removed :) thanks

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
```diff
@@ -0,0 +1,263 @@
...
+        .when(statsRecorderWrapperForConnection)
+        .putAndRecordPerConnectionErrorCount(errorCountCaptor.capture());
+    runInterceptorTasksAndAssertCount(1);
+    List<Long> allErrorCounts = errorCountCaptor.getAllValues();
+    assertThat(allErrorCounts.size()).isEqualTo(1);
+    assertThat(allErrorCounts.get(0)).isEqualTo(errorCount);
+  }
+
+  @Test
+  public void readWithTwoStubs() throws Exception {
```
mutianf, line 109: readWithTwoStubs -> readWithTwoChannels

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+  @After
+  public void tearDown() throws Exception {
+    if (server != null) {
+      server.shutdown();
+    }
+  }
+
+  @Test
+  public void singleRead() throws Exception {
```
mutianf, line 80: singleRead -> readWithOneChannel

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
```diff
@@ -0,0 +1,263 @@
...
+
+  @Test
+  public void noFailedRequests() throws Exception {
+    EnhancedBigtableStub stub = EnhancedBigtableStub.create(builder.build());
+
+    for (int i = 0; i < 20; i++) {
+      try {
+        stub.readRowsCallable().call(Query.create(SUCCESS_TABLE_NAME)).iterator().hasNext();
+      } catch (Exception e) {
+        // noop
```
mutianf, line 218: maybe remove the try catch since it's not necessary?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
```diff
@@ -0,0 +1,263 @@
...
+    Mockito.doNothing()
+        .when(statsRecorderWrapperForConnection)
+        .putAndRecordPerConnectionErrorCount(errorCountCaptor.capture());
+    runInterceptorTasksAndAssertCount(1);
+    List<Long> allErrorCounts = errorCountCaptor.getAllValues();
+    assertThat(allErrorCounts.size()).isEqualTo(1);
+    assertThat(allErrorCounts.get(0)).isEqualTo(0);
+  }
+
+  private void runInterceptorTasksAndAssertCount(int expectNumOfTasks) {
```
mutianf, line 231: now we're only creating 1 stub for all the test, is this step still necessary?

mutianf (changes requested):

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
rkaregar: I tried without it and the test fails because of the error that we're introducing

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/ErrorCountPerConnectionTest.java
rkaregar: I cleaned up the function to remove the argument, now we just check to make sure the number of tasks created is always 1.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ErrorCountPerConnectionMetricTracker.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.common.collect.ImmutableMap;
+import io.grpc.ClientInterceptor;
+import java.util.Collections;
+import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+/* Background task that goes through all connections and updates the errors_per_connection metric. */
+public class ErrorCountPerConnectionMetricTracker implements Runnable {
```
igorbernstein2, line 31: This should be marked Internal

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ErrorCountPerConnectionMetricTracker.java
```diff
@@ -0,0 +1,88 @@
...
+  private static final Integer PER_CONNECTION_ERROR_COUNT_PERIOD_SECONDS = 60;
+  private final EnhancedBigtableStubSettings.Builder builder;
```
igorbernstein2, line 33: Please dont retain long running references to Builders....Builders are supposed to be shortlived data carriers. Also passing in the entire settings Builder for just single value makes things hard to follow

Finally `getBackgroundExecutorProvider().getExecutor();` can and will return different instances of the an executor provider on every call.

You need to pass in the actual instance of the ScheduledExecutor that CLientContext will contain

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/ErrorCountPerConnectionTest.java
```diff
@@ -0,0 +1,265 @@
...
+import com.google.api.gax.grpc.ChannelPoolSettings;
+import com.google.api.gax.grpc.InstantiatingGrpcChannelProvider;
+import com.google.bigtable.v2.*;
+import com.google.cloud.bigtable.data.v2.BigtableDataSettings;
+import com.google.cloud.bigtable.data.v2.FakeServiceBuilder;
+import com.google.cloud.bigtable.data.v2.models.*;
+import com.google.cloud.bigtable.data.v2.stub.EnhancedBigtableStub;
+import com.google.cloud.bigtable.data.v2.stub.EnhancedBigtableStubSettings;
+import com.google.cloud.bigtable.stats.StatsRecorderWrapperForConnection;
+import io.grpc.*;
```
igorbernstein2, line 32: please remove all wildcard imports

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -204,10 +205,26 @@ public static ClientContext createClientContext(EnhancedBigtableStubSettings set
             ? ((InstantiatingGrpcChannelProvider) builder.getTransportChannelProvider()).toBuilder()
             : null;
 
-    if (builder.getEnableRoutingCookie() && transportProvider != null) {
-      // TODO: this also need to be added to BigtableClientFactory
-      // patch cookies interceptor
-      transportProvider.setInterceptorProvider(() -> ImmutableList.of(new CookiesInterceptor()));
+    ErrorCountPerConnectionMetricTracker errorCountPerConnectionMetricTracker =
+        new ErrorCountPerConnectionMetricTracker(
+            builder.getBackgroundExecutorProvider().getExecutor(),
```
mutianf, line 210: I think we want to get the executor from the client context. We can have your ` startConnectionErrorCountTracker` takes in a executor and start the task after the client context is built. 

So on line 253, we can have:

ClientContext clientContext = ClientContext.create(builder.build());
errorCountPerConnectionMetricTracker.startConnectionErrorCountTracker(clientContext.getExecutor());

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -204,10 +205,26 @@ public static ClientContext createClientContext(EnhancedBigtableStubSettings set
...
+            createBuiltinAttributes(builder));
```
igorbernstein2: This should probably be only instantiated if you can set it on the transportProvider.

Also `builder.getBackgroundExecutorProvider().getExecutor()` will have the same problem here as I described in my previous comment...it will return a different instance.

You have 2 options:

1. You can create your own fixed executorProvider (similar to FixedExecutorProvider, but respects close())
2. you can break the circular dependency on the executor by moving it to the start() method of the tracker and call it after the client context is constructed

igorbernstein2 (changes requested):

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/ErrorCountPerConnectionTest.java
```diff
@@ -0,0 +1,267 @@
...
+        runnable.run();
+        actualNumOfTasks++;
+      }
+    }
+    assertThat(actualNumOfTasks).isEqualTo(1);
+  }
+
+  static class FakeService extends BigtableGrpc.BigtableImplBase {
+
+    private final AtomicInteger count = new AtomicInteger();
```
mutianf, line 250: nit: is this used?

mutianf (approved):
lgtm after the final nit

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -254,13 +271,7 @@ public static ApiTracerFactory createBigtableTracerFactory(
...
             .put(RpcMeasureConstants.BIGTABLE_APP_PROFILE_ID, TagValue.create(appProfileId))
             .build();
-    ImmutableMap<String, String> builtinAttributes =
-        ImmutableMap.<String, String>builder()
-            .put("project_id", projectId)
-            .put("instance", instanceId)
-            .put("app_profile", appProfileId)
-            .put("client_name", "bigtable-java/" + Version.VERSION)
-            .build();
+    ImmutableMap<String, String> builtinAttributes = createBuiltinAttributes(settings.toBuilder());
```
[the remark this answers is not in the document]
rkaregar: Here we're just replacing the previous value assignment with a new function, so I think it wouldn't create staleness if it didn't already exist?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
rkaregar: Went with approach 2

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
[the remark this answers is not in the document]
igorbernstein2: oi, nevermind, brain is not working well

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ConnectionErrorCountInterceptor.java
```diff
@@ -0,0 +1,74 @@
...
+                try {
+                  if (status.isOk()) {
+                    numOfSuccesses.increment();
+                  } else {
+                    numOfErrors.increment();
+                  }
+                } finally {
+                  super.onClose(status, trailers);
+                }
+              }
```
igorbernstein2, line 60: I'm a little concerned about both throwing the exception and invoking super.onClose.

It would be safer to do this:

```java
public void onClose(Status status, Metadata trailers) {
     // connection accounting is non-critical so we log the exception, but let normal processing proceed
   try {
     handleOnCloseUnsafe(status, trailers);
  } catch (Throwable t) {
    if (t instanceof InterruptedException) {
       Thread.current().interrupt();
     }
    LOG.exception("Failed to record connection stats", e);
  }
    super.onClose(status, trailers);
}
private void handleOnCloseUnsafe(Status status, Metadata trailers) {
 if (status.isOk()) {
    numOfSuccesses.increment();
  } else {
    numOfErrors.increment();
  }
}

```

igorbernstein2 (approved):

PR #2133, opened 2024-02-26
feat: migrate per connection error count metric to otel

mutianf:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
@@ -16,6 +16,7 @@
 package com.google.cloud.bigtable.data.v2.stub;
 
 import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.APP_PROFILE_KEY;
+import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.CLIENT_NAME_KEY;
 import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.INSTANCE_ID_KEY;
 import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.PROJECT_ID_KEY;
 
@@ -99,12 +100,11 @@ import com.google.cloud.bigtable.data.v2.stub.changestream.ReadChangeStreamUserC
 import com.google.cloud.bigtable.data.v2.stub.metrics.BigtableTracerStreamingCallable;
 import com.google.cloud.bigtable.data.v2.stub.metrics.BigtableTracerUnaryCallable;
 import com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsTracerFactory;
-import com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsView;
 import com.google.cloud.bigtable.data.v2.stub.metrics.CompositeTracerFactory;
 import com.google.cloud.bigtable.data.v2.stub.metrics.CustomOpenTelemetryMetricsProvider;
 import com.google.cloud.bigtable.data.v2.stub.metrics.DefaultMetricsProvider;
-import com.google.cloud.bigtable.data.v2.stub.metrics.MetricsProvider;
 import com.google.cloud.bigtable.data.v2.stub.metrics.ErrorCountPerConnectionMetricTracker;
+import com.google.cloud.bigtable.data.v2.stub.metrics.MetricsProvider;
 import com.google.cloud.bigtable.data.v2.stub.metrics.MetricsTracerFactory;
 import com.google.cloud.bigtable.data.v2.stub.metrics.NoopMetricsProvider;
 import com.google.cloud.bigtable.data.v2.stub.metrics.RpcMeasureConstants;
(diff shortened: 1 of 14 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -323,53 +325,47 @@ public static ApiTracerFactory createBigtableTracerFactory(
...
   }
 
-  private static BuiltinMetricsTracerFactory createBuiltinMetricsTracerFactory(
-      String projectId,
-      MetricsProvider metricsProvider,
-      Attributes attributes,
-      ClientContext clientContext)
+  @Nullable
+  public static OpenTelemetry getOpenTelemetry(
+      String projectId, MetricsProvider metricsProvider, Credentials defaultCredentials)
```
igorbernstein2, line 340: Please mark Credentials as Nullable.

Also is the current behavior is intentional? ie NoCredentialProvider will return null Credentials, but that will be passed down into the metrics client as FixedCredentials (null)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsView.java
```diff
@@ -26,7 +26,7 @@
 import java.util.Map;
 
 /**
- * A util class to register built-in metrics on a custom OpenTelemetry instance. This is for
+ * An util class to register built-in metrics on a custom OpenTelemetry instance. This is for
```
igorbernstein2, line 29: I think `A` was correct :)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/ErrorCountPerConnectionMetricTracker.java
```diff
@@ -42,12 +52,25 @@ void setStatsRecorderWrapperForConnection(
...
 
-  public ErrorCountPerConnectionMetricTracker(ImmutableMap<String, String> builtinAttributes) {
+  public ErrorCountPerConnectionMetricTracker(
+      @Nullable OpenTelemetry openTelemetry, Attributes attributes) {
     connectionErrorCountInterceptors =
         Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()));
 
-    this.statsRecorderWrapperForConnection =
-        StatsWrapper.createRecorderForConnection(builtinAttributes);
+    if (openTelemetry != null) {
```
igorbernstein2, line 60: Why even create this class if opentelemetry is null?

igorbernstein2 (changes requested):

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/BigtableDataClientFactoryTest.java
```diff
@@ -320,7 +320,7 @@ public void testFeatureFlags() throws Exception {
   @Test
   public void testBulkMutationFlowControllerConfigured() throws Exception {
     BigtableDataSettings settings =
-        BigtableDataSettings.newBuilder()
+        BigtableDataSettings.newBuilderForEmulator(server.getPort())
```
rkaregar, line 323: Was this a bug or there's a reason that we now have to use the emulator in this test?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/ErrorCountPerConnectionTest.java
```diff
@@ -131,28 +181,42 @@ public void readWithTwoChannels() throws Exception {
...
+            allMetrics, BuiltinMetricsConstants.PER_CONNECTION_ERROR_COUNT_NAME);
+
+    // The 2 channels should get equal amount of errors, so the totalErrorCount / 2 bucket is
+    // updated twice.
+    ArrayList<HistogramPointData> histogramPointData =
+        new ArrayList<>(metricData.getHistogramData().getPoints());
+    assertThat(histogramPointData.size()).isEqualTo(1);
+    HistogramPointData point = histogramPointData.get(0);
+    int index = 0;
+    for (Double boundary : point.getBoundaries()) {
```
rkaregar, line 200: nit: maybe we can extract this into a function that returns the first index where the boundary is larger than the error count

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/ErrorCountPerConnectionTest.java
```diff
@@ -162,52 +226,48 @@ public void readOverTwoPeriods() throws Exception {
...
     runInterceptorTasksAndAssertCount();
-    allErrorCounts = errorCountCaptor.getAllValues();
-    assertThat(allErrorCounts.size()).isEqualTo(1);
-    assertThat(allErrorCounts.get(0)).isEqualTo(errorCount);
-  }
 
-  @Test
-  public void ignoreInactiveConnection() throws Exception {
-    EnhancedBigtableStub stub = EnhancedBigtableStub.create(builder.build());
+    Collection<MetricData> allMetrics = metricReader.collectAllMetrics();
```
rkaregar, line 250: so the points recorded using `perConnectionErrorCountHistogram.record()` stay in the metricReader until `metricReader.collectAllMetrics()` is called and they're picked up? 
Since we're using a 60-second cycle for both collecting and exporting the metrics, I wonder if it's possible that on some minutes we could skip exporting metrics and end up sending them twice over the next minute. For example, we start the collection at time 59.998s and finish at 60.001s, but export at 60.000s. Next minute, we start the export at 199.998s until 199.999s and export at 120.000s, double-exporting the metrics. (I guess if it is indeed an issue, it would've also been the case before migrating to otel?)

mutianf: The StatsWrapper was introduced because we want a fixed OpenCensus version and namespace, and it's not a required abstraction once we migrated to OTEL. 

In this test, we use a in-memory metric reader which let us investigate the actual aggregated values so we don't need to use a mock anymore. We still trigger the task run in `runInterceptorTasksAndAssertCount` method. The issue you described could be a problem, but yes, I think it would exist before the migration.

igorbernstein2 (approved):

PR #2143, opened 2024-03-04
fix: per-connection metrics issue when using a different Bigtable project

rkaregar:
Fix an issue in exporting per-connection metrics when using a different Bigtable project.


google-cloud-bigtable-stats/src/main/java/com/google/cloud/bigtable/stats/BigtableStackdriverExportUtils.java
```diff
@@ -55,6 +55,11 @@
 
 class BigtableStackdriverExportUtils {
   private static final String BIGTABLE_RESOURCE_TYPE = "bigtable_client_raw";
+
+  // Defined package private to access in tests.
```
mutianf, line 59: Instead of the comment you can annotate with `@VisibleForTesting`

mutianf (approved):
lgtm after nit.

PR #2151, opened 2024-03-07
feat: Add bigtable admin CRUD methods for AuthorizedView

trollyxia:
Change-Id: Ie6b2881d99d5f787656df04f2d7de98e59bb9642


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -264,7 +331,7 @@ public boolean exists(String tableId, ByteString rowKey) {
    * }</pre>
    */
   public ApiFuture<Boolean> existsAsync(String tableId, String rowKey) {
-    return existsAsync(tableId, ByteString.copyFromUtf8(rowKey));
+    return existsAsync(tableId, rowKey, new ExistsOptions());
```
kolea2, line 334: why are we removing `ByteString.copyFromUtf8(rowKey)` here?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -370,8 +505,7 @@ public Row readRow(String tableId, ByteString rowKey) {
    * @throws com.google.api.gax.rpc.ApiException when a serverside error occurs
    */
   public Row readRow(String tableId, String rowKey) {
-    return ApiExceptions.callAndTranslateApiException(
-        readRowAsync(tableId, ByteString.copyFromUtf8(rowKey), null));
+    return readRow(tableId, rowKey, new ReadRowOptions());
```
kolea2, line 508: same question here

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableAuthorizedViewIT.java
```diff
@@ -0,0 +1,257 @@
+/*
+ * Copyright 2023 Google LLC
```
kolea2, line 2: here and everywhere else, 2024

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableAuthorizedViewIT.java
```diff
@@ -0,0 +1,257 @@
...
+      fail("AuthorizedView was not deleted.");
+    } catch (NotFoundException e) {
+      assertWithMessage("Incorrect exception type")
+          .that(e.getCause())
+          .isInstanceOf(StatusRuntimeException.class);
+    }
+  }
+
+  @Test
+  public void AuthorizedViewIamTest() {
```
kolea2, line 200:
```suggestion
  public void authorizedViewIamTest() {
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+      return Objects.hashCode(builder.build());
+    }
+  }
+
+  /** Represents subsets of a particular column family that are included in this authorized view. */
+  public static final class FamilySubsets {
```
kolea2, line 219: can we separate each of the `public static final class`es into their own files? This is making this one file really large.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
```diff
@@ -0,0 +1,309 @@
...
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(proto);
+  }
+
+  /**
+   * Represents a subset of a Table. Please check the implementations of this interface for more
+   * details.
+   */
+  public interface AuthorizedViewImpl {}
```
kolea2, line 125: is this necessary?  If so, are we able to mark this as `@InternalApi` (thinking of backwards compatibility)?

kolea2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
kolea2: also, if necessary, can we remove the `Impl` since it's the interface?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -208,7 +213,7 @@ static BigtableDataClient createWithClientContext(
    * @throws com.google.api.gax.rpc.ApiException when a serverside error occurs
    */
   public boolean exists(String tableId, String rowKey) {
-    return ApiExceptions.callAndTranslateApiException(existsAsync(tableId, rowKey));
+    return exists(tableId, rowKey, new ExistsOptions());
```
mutianf, line 216: Here and everywhere, Instead of `new ExistsOptions()`, can we have something like `ExistsOptions.create()` so it's consistent with the rest of the API? And remember to update the javadoc after you make the change. And when we want to set ExistOptions for a view id, we can do `ExistsOptions.newBuilder().setAuthorizedViewId(..).build`, or simply `ExistsOptions.createForAuthotizedView(..)`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
mutianf: I think this will call `public Row readRow(String tableId, String rowKey, @Nullable ReadRowOptions readRowOptions) {` and the rowKey will be converted later

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
mutianf: I think this will call:
```
  public ApiFuture<Boolean> existsAsync(
      String tableId, String rowKey, @Nullable ExistsOptions existsOptions) {
    return existsAsync(tableId, ByteString.copyFromUtf8(rowKey), existsOptions);
```
which the string is converted there

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -880,7 +1172,39 @@ public <RowT> ServerStreamingCallable<Query, RowT> readRowsCallable(RowAdapter<R
    * @throws com.google.api.gax.rpc.ApiException when a serverside error occurs
    */
   public List<KeyOffset> sampleRowKeys(String tableId) {
-    return ApiExceptions.callAndTranslateApiException(sampleRowKeysAsync(tableId));
+    return sampleRowKeys(tableId, new SampleRowKeysOptions());
```
mutianf, line 1175: same for `new SampleRowKeysOptions()`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -235,7 +240,69 @@ public boolean exists(String tableId, String rowKey) {
...
+   *     System.out.println(key + " is present");
+   *   }
+   * } catch(ApiException e) {
+   *   e.printStackTrace();
+   * }
+   * }</pre>
+   *
+   * @throws com.google.api.gax.rpc.ApiException when a serverside error occurs
+   */
+  public boolean exists(String tableId, String rowKey, @Nullable ExistsOptions existsOptions) {
```
mutianf, line 273: Here and everywhere, do we need the `@Nullable` annotation here? looks like other methods are calling with an empty instance of the existOption?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ReadRowOptions.java
```diff
@@ -0,0 +1,59 @@
+/*
+ * Copyright 2023 Google LLC
```
mutianf, line 2: Here and everywhere in the new classes, 2023 -> 2024

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ReadModifyWriteRow.java
```diff
@@ -50,10 +58,33 @@ public static ReadModifyWriteRow create(@Nonnull String tableId, @Nonnull String
...
   }
 
+  /**
+   * Creates a new instance of the ReadModifyWriteRow on an authorized view.
+   *
+   * <p>See {@link com.google.cloud.bigtable.admin.v2.models.AuthorizedView} for more details about
+   * an AuthorizedView.
+   */
+  public static ReadModifyWriteRow createForAuthorizedView(
+      @Nonnull String tableId, @Nonnull String authorizedViewId, @Nonnull String key) {
```
mutianf, line 68: Preconditions.checkNotNull(authorizedViewId, "authorized view id can't be null");

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutation.java
```diff
@@ -72,6 +108,31 @@ public static RowMutation create(
...
+   * Creates new instance of mutation builder on an authorized view by wrapping existing set of row
+   * mutations. The builder will be owned by this RowMutation and should not be used by the caller
+   * after this call. This functionality is intended for advanced usage.
+   *
+   * <p>Sample code:
+   *
+   * <pre><code>
+   * Mutation mutation = Mutation.create()
+   *     .setCell("[FAMILY_NAME]", "[QUALIFIER]", [TIMESTAMP], "[VALUE]");
+   * RowMutation rowMutation = RowMutation.create("[TABLE]", "[AUTHORIZED_VIEW]", "[ROW_KEY]", mutation);
```
mutianf, line 121: RowMutation.create -> RowMutation.createForAuthorizedVie

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutation.java
```diff
@@ -90,6 +151,30 @@ public static RowMutation create(
...
+   * Creates new instance of mutation builder on an authorized view by wrapping existing set of row
+   * mutations. The builder will be owned by this RowMutation and should not be used by the caller
+   * after this call. This functionality is intended for advanced usage.
+   *
+   * <p>Sample code:
+   *
+   * <pre><code>
+   * Mutation mutation = Mutation.create()
+   *     .setCell("[FAMILY_NAME]", "[QUALIFIER]", [TIMESTAMP], "[VALUE]");
+   * RowMutation rowMutation = RowMutation.create("[TABLE]", "[AUTHORIZED_VIEW]", [BYTE_STRING_ROW_KEY], mutation);
```
mutianf, line 164: RowMutation.create -> RowMutation.createForAuthorizedView

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -811,6 +902,37 @@ public Batcher<RowMutationEntry, Void> newMutateRowsBatcher(
...
+   *       been applied, are filtered from the next attempt. Also, any entries that failed with a
+   *       nontransient error, are filtered from the next attempt. This will continue until there
+   *       are no more entries or there are no more retry attempts left.
+   *   <li>Wrap batch failures in a {@link
+   *       com.google.cloud.bigtable.data.v2.models.MutateRowsException}.
+   *   <li>Split the responses using {@link MutateRowsBatchingDescriptor}.
+   * </ul>
+   */
+  public Batcher<RowMutationEntry, Void> newMutateRowsBatcher(
+      @Nonnull BulkMutation mutation, @Nullable GrpcCallContext ctx) {
```
mutianf, line 925: why do we need this API? the batcher is suppose to group RowMutationEntries into BulkMutations. Maybe instead we can do newMutateRowsBatcher(tableId, viewId, ctx)?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -1166,6 +1303,10 @@ public UnaryCallable<String, List<KeyOffset>> sampleRowKeysCallable() {
     return sampleRowKeysCallable;
   }
 
+  public UnaryCallable<SampleRowKeys, List<KeyOffset>> sampleRowKeysCallable2() {
```
mutianf, line 1306: maybe rename it to `sampleRowKeysCallableWithRequest()` or something?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/internal/NameUtil.java
```diff
@@ -55,6 +58,11 @@ public static String formatBackupName(
     return formatClusterName(projectId, instanceId, clusterId) + "/backups/" + backupId;
   }
 
+  public static String formatAuthorizedViewName(
```
kolea2, line 61: this needs a test in `NameUtilTest.java`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
```diff
@@ -0,0 +1,309 @@
...
+    return fullName.getTable();
+  }
+
+  /** Returns whether this authorized view is deletion protected. */
+  public boolean isDeletionProtected() {
+    return proto.getDeletionProtection();
+  }
+
+  /** Gets the implementation of this authorized view, which currently can only be a subset view. */
+  public AuthorizedViewImpl getAuthorizedViewImpl() {
```
kolea2, line 86: is this only used for testing? Please annotate with `@VisibleForTesting`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/NameUtil.java
```diff
@@ -40,11 +42,46 @@ public static String formatTableName(
...
+
   public static String extractTableIdFromTableName(@Nonnull String fullTableName) {
     Matcher matcher = TABLE_PATTERN.matcher(fullTableName);
     if (!matcher.matches()) {
       throw new IllegalArgumentException("Invalid table name: " + fullTableName);
     }
     return matcher.group(3);
   }
+
+  public static String extractTableIdFromAuthorizedViewName(
```
kolea2, line 61: same here, these also need tests.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedViewTest.java
```diff
@@ -0,0 +1,135 @@
...
+import com.google.cloud.bigtable.admin.v2.models.AuthorizedView.FamilySubsets;
+import com.google.cloud.bigtable.admin.v2.models.AuthorizedView.SubsetView;
+import com.google.protobuf.ByteString;
+import java.util.Map;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class AuthorizedViewTest {
```
kolea2, line 31: equals and hashcode should also be tested here (same goes for other newly created models)

kolea2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
trollyxia: Hmm these classes are only useful in the context of AuthorizedViews so I would prefer keeping those in the `AuthorizedView.java`.
Also given the urgency of this PR, can we keep it this way for now, since we can always do the refactor later?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
trollyxia: Right now there is only one flavour of AuthorizedView, which is the SubsetView class. But there might be more flavours in the future and we would like to keep the interface extensible, hence the `AuthorizedViewImpl` interface.

I renamed it to `AuthorizedViewType` instead. Let me know if you have better idea.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClient.java
```diff
@@ -1434,6 +1442,352 @@ public ApiFuture<Void> awaitReplicationAsync(final String tableId) {
...
+   * Asynchronously creates a new authorized view with the specified configuration.
+   *
+   * <p>Sample code:
+   *
+   * <pre>{@code
+   * CreateAuthorizedViewRequest request =
+   *     CreateAuthorizedViewRequest.of("my-table", "my-new-authorized-view")
+   *         .setDeletionProtection(true)
+   *         .setAuthorizedViewImpl(
+   *             new AuthorizedView.SubsetView()
```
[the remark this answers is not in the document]
trollyxia: Hmmm... I think the `of()` call without any parameter here is a bit off the paradigm? Normally we only do things like `CreateTableRequest.of("my-table")`, but the SubsetView here is not strictly belong to something, so the `of()` call doesn't quite apply to it?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
trollyxia: Done. Please take another look.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
trollyxia: I suppose we can do newMutateRowsBatcher(tableId, viewId, ctx).

But similar to the `exists()` and everything else, we didn't simply add a `viewId` argument to the existing function call but added a new `ExistsOption` argument. I think by doing this we avoid the potential growth of the arguments when adding of new features in the future, since we can simply add a new field in the options class rather than add a new API with more arguments.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
trollyxia: I think potentially this annotation won't do any harm as well?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
trollyxia: I think we do want this to be public so that users can inspect what's in their authorized views.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
kolea2: sg!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
kolea2: moving the classes would be considered a breaking change, so we should decide now. @mutianf do you have a preference?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
trollyxia: I think this @Nullable annotation enables users to call this function like `exists("my-table", "row1", nullptr)`. Let me know if you don't think we should allow this type of usage and I'll remove the annotation.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/MutateRowOptions.java
```diff
@@ -0,0 +1,60 @@
...
+    return new MutateRowOptions();
+  }
+
+  /**
+   * Limits the mutation to subsets of the table represented in an authorized view with the
+   * specified authorized view id.
+   *
+   * @see com.google.cloud.bigtable.admin.v2.models.AuthorizedView for more details.
+   */
+  public static MutateRowOptions createForAuthorizedView(@Nullable String authorizedViewId) {
```
mutianf, line 41: authorizedViewId should not be null if user is calling `createForAuthorizedView`. They could call create() instead if they don't want authorizedViewId to be set. And we should add a `Precondition.checkNotNull(authorizedViewId, "use create() instead"`.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/MutateRowOptions.java
```diff
@@ -0,0 +1,60 @@
...
+    return new MutateRowOptions(authorizedViewId);
+  }
+
+  private MutateRowOptions() {}
+
+  private MutateRowOptions(@Nonnull String authorizedViewId) {
```
mutianf, line 47: You can mark authorizedViewId as Nullable here, and we don't need the MutateRowOptions() constructor. In create(), we can just call `return new MutateRowOptions(null)`.

The `getAuthorizedViewId` should also be annotated with `@Nullable`.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/MutateRowOptions.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.common.base.Preconditions;
+import java.io.Serializable;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * Represents a collection of customized options for mutating a row. Currently, it only wraps an
+ * authorized view id option but there could be more in the future.
+ */
+public class MutateRowOptions implements Serializable {
```
mutianf, line 28: You can use AutoValue for all the Options class. Something like:

```
@AutoValue
public abstract class Options {

    @Nullable private String viewId;

    public static Options create() {
        return new AutoValue_Options(null);
    }

    public static Options createForView(String view) {
        return new AutoValue_Options(view);
    }

    @Nullable abstract String getViewId();
}
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
mutianf: Maybe we can move SubsetView and FamilySubset out to a new class? WE might add more AuthorizedViewTypes in the future, so maybe it's better to move them out now and avoid this file get even larger ?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutation.java
```diff
@@ -49,11 +63,33 @@ public static RowMutation create(@Nonnull String tableId, @Nonnull String key) {
...
   }
 
+  /**
+   * Creates a new instance of the mutation builder on an authorized view.
+   *
+   * <p>See {@link com.google.cloud.bigtable.admin.v2.models.AuthorizedView} for more details about
+   * an AuthorizedView.
+   */
+  public static RowMutation createForAuthorizedView(
+      @Nonnull String tableId, @Nonnull String authorizedViewId, @Nonnull String key) {
```
mutianf, line 73: `Precondition.checkNotNull(authorizedViewId);`

The `@Nonnull` annotation is for code mantainers to know the expected object value. it doesn't really enforce anything.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutation.java
```diff
@@ -49,11 +63,33 @@ public static RowMutation create(@Nonnull String tableId, @Nonnull String key) {
...
   }
 
+  /**
+   * Creates a new instance of the mutation builder on an authorized view.
+   *
+   * <p>See {@link com.google.cloud.bigtable.admin.v2.models.AuthorizedView} for more details about
+   * an AuthorizedView.
+   */
+  public static RowMutation createForAuthorizedView(
+      @Nonnull String tableId, @Nonnull String authorizedViewId, @Nonnull ByteString key) {
```
mutianf, line 89: `Precondition.checkNotNull(authorizedViewId);`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/SampleRowKeys.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Creates a new instance of the sample row keys builder on an authorized view.
+   *
+   * <p>See {@link com.google.cloud.bigtable.admin.v2.models.AuthorizedView} for more details about
+   * an AuthorizedView.
+   */
+  public static SampleRowKeys createForAuthorizedView(
+      @Nonnull String tableId, @Nonnull String authorizedViewId) {
```
mutianf, line 56: `Precondition.checkNotNull(authorizedViewId);`

mutianf (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
trollyxia: Sure. Just to confirm, do you mean to move the SubsetView and FamilySubset to two separate files?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
trollyxia: This does allows the users to call the API in a uniformed manner though, regardless of whether they have an initialized `ExistsOptions` or it's a `nullptr`. Therefore, users can simply call this API without needing to differentiate the two cases with an `if` statement.

Also I found that we have similar APIs with the `Nullable` annotation like the following:
```
public Row readRow(String tableId, String rowKey, @Nullable Filter filter) {
    return readRow(tableId, rowKey, ReadRowOptions.create().filter(filter));
}
```

What do you think?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/MutateRowOptions.java
trollyxia: Somehow the code doesn't compile for me.

My code:

```
@AutoValue
public abstract class ExistsOptions implements Serializable {
  @Nullable private String authorizedViewId;

  public static ExistsOptions create() {
    return new AutoValue_ExistsOptions(null);
  }

  public static ExistsOptions createForAuthorizedView(String authorizedViewId) {
    return new AutoValue_ExistsOptions(authorizedViewId);
  }

  @Nullable
  public String getAuthorizedViewId() {
    return this.authorizedViewId;
  }
}
```
And the error was `AutoValue_ExistsOptions() Expected 0 arguments but found 1`.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
mutianf: But I think the caller should differentiate the null vs has a viewId case? Because we do expect different results for the 2 cases?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
```diff
@@ -48,10 +50,33 @@ public static BulkMutation create(String tableId) {
...
+   * Creates a new instance of the bulk mutation builder on an AuthorizedView.
+   *
+   * <p>See {@link com.google.cloud.bigtable.admin.v2.models.AuthorizedView} for more details about
+   * an AuthorizedView.
+   */
+  public static BulkMutation createForAuthorizedView(String tableId, String authorizedViewId) {
+    return new BulkMutation(tableId, authorizedViewId);
+  }
+
+  @InternalApi("For internal use only")
```
mutianf, line 63: where is this used? Can it be package private?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/MutateRowsOptions.java
```diff
@@ -0,0 +1,51 @@
...
+    return new AutoValue_MutateRowsOptions(null);
+  }
+
+  /**
+   * Limits the mutations to subsets of the table represented in an authorized view with the
+   * specified authorized view id.
+   *
+   * @see com.google.cloud.bigtable.admin.v2.models.AuthorizedView for more details.
+   */
+  public static MutateRowsOptions createForAuthorizedView(@Nullable String authorizedViewId) {
```
mutianf, line 41: we shouldn't have the `@Nullable` annotation here

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
```diff
@@ -59,8 +61,64 @@ public static Query create(String tableId) {
...
+   * com.google.cloud.bigtable.data.v2.BigtableDataSettings}.
+   *
+   * <p>See {@link com.google.cloud.bigtable.admin.v2.models.AuthorizedView} for more details about
+   * an AuthorizedView.
+   */
+  public static Query createForAuthorizedView(String tableId, String authorizedViewId) {
+    return new Query(tableId, authorizedViewId);
+  }
+
+  @InternalApi("For internal use only")
```
mutianf, line 77: same question here, where is this used? could it be pacakge private?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
```diff
@@ -59,8 +61,64 @@ public static Query create(String tableId) {
...
+  public static Query create(String tableId, @Nullable ExistsOptions existsOptions) {
```
mutianf, line 78: why do we having ExistsOptions, ReadRowOptions and ReadRowsOptions? for query I think we should only allow ReadRowsOptions? especially if it's internal use only, the caller should be able to call createForAuthorizedView instead?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ReadRowOptions.java
```diff
@@ -0,0 +1,68 @@
...
+    return new ReadRowOptions(null);
+  }
+
+  /**
+   * Limits the read to subsets of the table represented in an authorized view with the specified
+   * authorized view id.
+   *
+   * @see com.google.cloud.bigtable.admin.v2.models.AuthorizedView for more details.
+   */
+  public static ReadRowOptions createForAuthorizedView(@Nullable String authorizedViewId) {
```
mutianf, line 44: same here, we shouldn't have the annotation here

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/ReadRowsOptions.java
```diff
@@ -0,0 +1,68 @@
...
+    return new ReadRowsOptions(null);
+  }
+
+  /**
+   * Limits the reads to subsets of the table represented in an authorized view with the specified
+   * authorized view id.
+   *
+   * @see com.google.cloud.bigtable.admin.v2.models.AuthorizedView for more details.
+   */
+  public static ReadRowsOptions createForAuthorizedView(@Nullable String authorizedViewId) {
```
mutianf, line 44: same here, let's remove the annotation

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
trollyxia: It's used here: https://github.com/googleapis/java-bigtable/pull/2151/files#diff-0bda21f28dd389fa7bff1f50b603396b0e643b4cd1c98ce8b6a3ec3fce0b307fR1595-R1600 so it cannot be private?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
trollyxia: They are used in different places in BigtableDataClient.java, to convert a `ExistsOptions`/`ReadRowOptions`/`ReadRowsOptions` to a Query conveniently.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
trollyxia: Done.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -367,6 +370,7 @@ public EnhancedBigtableStub(
     readRowCallable = createReadRowCallable(new DefaultRowAdapter());
     bulkReadRowsCallable = createBulkReadRowsCallable(new DefaultRowAdapter());
     sampleRowKeysCallable = createSampleRowKeysCallable();
+    sampleRowKeysCallable2 = createSampleRowKeysCallableWithRequest();
```
mutianf, line 373: can we rename this `sampleRowKeysCallable2` as well?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
mutianf: The javadoc you referenced is not using this method?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
mutianf: Query should always use ReadRowsOptions. Can you point me to the code where other Options are used?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
trollyxia: If you check the implementation of the current `newBulkReadRowsBatcher(
      String tableId, @Nullable Filters.Filter filter, @Nullable GrpcCallContext ctx)`[link](https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java#L1260-L1267):

```
public Batcher<ByteString, Row> newBulkReadRowsBatcher(
      String tableId, @Nullable Filters.Filter filter, @Nullable GrpcCallContext ctx) {
    Query query = Query.create(tableId);
    if (filter != null) {
      query.filter(filter);
    }
    return stub.newBulkReadRowsBatcher(query, ctx);
  }

```
It didn't use the `filter` argument directly. Instead, it converts the `filter` into a `Query` object and calls the `newBulkReadRowsBatcher()` on that query.

After this change, the `filter` argument will be wrapped in a `ReadRowsOptions` (which also wraps an authorized view id), and we still converts the `readRowsOptions` into a `Query` object first:

```
  public Batcher<ByteString, Row> newBulkReadRowsBatcher(
      String tableId, @Nonnull ReadRowsOptions readRowsOptions, @Nullable GrpcCallContext ctx) {
    Preconditions.checkNotNull(
        readRowsOptions, "Please use the newBulkReadRowsBatcher(tableId, ctx) API instead");
    Query query = Query.create(tableId, readRowsOptions);
    return stub.newBulkReadRowsBatcher(query, ctx);
  }
```

In terms of the mutation, in my opinion, the mutation-counterpart for a `Query` object is a `bulkMutation`, which is why I choosed to convert a `MutateRowOptions` into a `bulkMutation`, and then calls the `newMutateRowsBatcher()` on that `bulkMutation`. I think this way the bulk mutation implementation would be more consistent to the bulk readrows implementation.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
trollyxia: Sorry about the broken link. This is related to the other discussion that we had, about converting a `MutateRowsOptions` into a `BulkMutation` first, and then calls the `newMutateRowsBatcher()` on the `BulkMutation` object. And you proposed to call the `newMutateRowsBatcher()` on the `MutateRowsOptions` directly. Let's discuss in the other comment.

If we have decided that we should call `newMutateRowsBatcher()` on the `MutateRowsOptions` directly, then I'll remove this internal function.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
trollyxia: Let's close this one and discuss in the comment below.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
trollyxia: The current `exists()` implementation actually builds a `Query` object:
https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java#L295-L314

Same thing for the `readRow()` method: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java#L586-L592

That's why we convert those options into a `Query`, in existsAsync() and readRowAsync().

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -954,6 +1334,49 @@ public UnaryCallable<String, List<KeyOffset>> sampleRowKeysCallable() {
...
+   *       }
+   *     }
+   *     public void onSuccess(List<KeyOffset> keyOffsets) {
+   *       System.out.println("Got key offsets: " + keyOffsets);
+   *     }
+   *   }, MoreExecutors.directExecutor());
+   * }
+   * }</pre>
+   */
+  public UnaryCallable<SampleRowKeys, List<KeyOffset>> sampleRowKeysCallable2() {
```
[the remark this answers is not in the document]
trollyxia: Done. Thanks for pointing this out!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
trollyxia: Also I wish to point out that even the original [newMutateRowsBatcher(
      @Nonnull String tableId, @Nullable GrpcCallContext ctx) implementation](https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java#L802-L812) actually created a `BulkMutation` object from the `tableId`.

I think in the end of the day, it is the `bulkMutateRowsCallable` requires a `BulkMutation` object, so we **have to** convert anything into a `BulkMutation` somewhere in the code.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Query.java
mutianf: Ah gotcha, I missed those. Thanks for clarifying!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(proto);
+  }
+
+  /**
+   * Represents a subset of a Table. Please check the implementations of this interface for more
+   * details.
+   */
+  public interface AuthorizedViewType {}
```
kolea2, line 121: I can't remember where we landed on this, but this should probably be marked as `@InternalApi` or `@InternalExtensionOnly`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/SubsetView.java
```diff
@@ -0,0 +1,121 @@
...
+    this.builder.addRowPrefixes(ByteString.copyFromUtf8(rowPrefix));
+    return this;
+  }
+
+  /**
+   * Adds a new familyId with its familySubsets to the subset view. Please note that calling this
+   * method with the same familyId will overwrite the previous rule set on the family.
+   */
+  public SubsetView addFamilySubsets(String familyId, FamilySubsets familySubsets) {
+    this.builder.putFamilySubsets(familyId, familySubsets.toProto());
```
kolea2, line 92: (assuming this is a map which is why it would be overwritten if you call this twice with the same familyId) could you instead do some logic here to append the new family subset if the key already exists?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
trollyxia: We have a `getAuthorizedViewType()` public method (as below) to allow users to inspect what's in their authorized views. This method returns an `AuthorizedViewType` object.

```
/** Gets the type of this authorized view, which currently can only be a subset view. */
public AuthorizedViewType getAuthorizedViewType() {
  if (proto.hasSubsetView()) {
    return SubsetView.fromProto(proto.getSubsetView());
  } else {
    // Should never happen because the constructor verifies that one must exist.
    throw new IllegalStateException("This AuthorizedView doesn't have a valid type specified");
  }
}
```

I'm not familiar with Java annotations. In this case, can it still be marked as `@InternalApi` or `@InternalExtensionOnly`?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/SubsetView.java
trollyxia: Great suggestion! Done.

PR #2160, opened 2024-03-08
feat: Add support for type and aggregate APIs

steveniemitz:
This PR adds support for type and aggregate APIs to the java client.  Bigtable aggregates will allow users to configure column families whose cells accumulate values via an aggregation function rather than simply overwrite them.

Change-Id: I5882459c862cc1deda11606e181cee38ac4d9099

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/TypeProtos.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/TypeProtos.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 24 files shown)


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/TypeProtos.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.bigtable.admin.v2;
+
+public class TypeProtos {
```
mutianf, line 18: `@InternalApi`? Is this only used in tests? `@VisibileForTesting`?

mutianf (approved):
lgtm after the nit

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/TypeProtos.java
steveniemitz: does it matter since this is already in the test module?

mutianf: Oh, nvm, I thought it's under src/, ignore my comment!

PR #2175, opened 2024-03-19
feat: Add admin APIs for AuthorizedView

trollyxia:
Change-Id: Ie31eae6da61ed0d0462e029f6247924785b239bf

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright 2024 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.cloud.bigtable.admin.v2.models;
+
+import com.google.api.core.InternalApi;
+import com.google.bigtable.admin.v2.AuthorizedViewName;
+import com.google.common.base.Objects;
+import com.google.common.base.Preconditions;
+import javax.annotation.Nonnull;
+
+/**
+ * A class that wraps the {@link com.google.bigtable.admin.v2.AuthorizedView} protocol buffer
+ * object.
+ *
+ * <p>An AuthorizedView represents subsets of a particular table based on rules. The access to each
+ * AuthorizedView can be configured separately from the Table.
+ *
+ * <p>Users can perform read/write operation on an AuthorizedView by providing an authorizedView id
+ * besides a table id, in which case the semantics remain identical as reading/writing on a Table
+ * except that visibility is restricted to the subset of the Table that the AuthorizedView
+ * represents.
+ */
+public final class AuthorizedView {
+  private final com.google.bigtable.admin.v2.AuthorizedView proto;
+
+  /**
+   * Wraps the protobuf. This method is considered an internal implementation detail and not meant
+   * to be used by applications.
+   */
+  @InternalApi
+  public static AuthorizedView fromProto(
+      @Nonnull com.google.bigtable.admin.v2.AuthorizedView proto) {
+    return new AuthorizedView(proto);
+  }
+
+  private AuthorizedView(@Nonnull com.google.bigtable.admin.v2.AuthorizedView proto) {
+    Preconditions.checkNotNull(proto);
+    Preconditions.checkArgument(!proto.getName().isEmpty(), "AuthorizedView must have a name");
+    Preconditions.checkArgument(
+        proto.hasSubsetView(), "AuthorizedView must have a subset_view field");
+    this.proto = proto;
+  }
+
+  /** Gets the authorized view's id. */
+  public String getId() {
+    // Constructor ensures that name is not null.
+    AuthorizedViewName fullName = AuthorizedViewName.parse(proto.getName());
+
+    //noinspection ConstantConditions
+    return fullName.getAuthorizedView();
+  }
+
+  /** Gets the id of the table that owns this authorized view. */
+  public String getTableId() {
+    // Constructor ensures that name is not null.
+    AuthorizedViewName fullName = AuthorizedViewName.parse(proto.getName());
+
+    //noinspection ConstantConditions
+    return fullName.getTable();
+  }
+
+  /** Returns whether this authorized view is deletion protected. */
+  public boolean isDeletionProtected() {
+    return proto.getDeletionProtection();
+  }
+
+  /** Gets the type of this authorized view, which currently can only be a subset view. */
+  public AuthorizedViewType getAuthorizedViewType() {
+    if (proto.hasSubsetView()) {
+      return SubsetView.fromProto(proto.getSubsetView());
+    } else {
+      // Should never happen because the constructor verifies that one must exist.
+      throw new IllegalStateException("This AuthorizedView doesn't have a valid type specified");
+    }
+  }
+
+  /**
+   * Creates the request protobuf. This method is considered an internal implementation detail and
+   * not meant to be used by applications.
+   */
+  @InternalApi
+  public com.google.bigtable.admin.v2.AuthorizedView toProto() {
+    return proto;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    AuthorizedView that = (AuthorizedView) o;
+    return Objects.equal(proto, that.proto);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(proto);
+  }
+
+  /**
+   * Represents a subset of a Table. Please check the implementations of this interface for more
+   * details.
+   */
+  public interface AuthorizedViewType {}
+}
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
@@ -17,6 +17,7 @@
 package com.google.cloud.bigtable.admin.v2.models;
 
 import com.google.api.core.InternalApi;
+import com.google.api.core.InternalExtensionOnly;
 import com.google.bigtable.admin.v2.AuthorizedViewName;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
@@ -118,6 +119,6 @@ public final class AuthorizedView {
    * Represents a subset of a Table. Please check the implementations of this interface for more
    * details.
    */
-  @InternalApi
+  @InternalExtensionOnly
   public interface AuthorizedViewType {}
 }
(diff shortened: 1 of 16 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
```diff
@@ -0,0 +1,122 @@
...
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(proto);
+  }
+
+  /**
+   * Represents a subset of a Table. Please check the implementations of this interface for more
+   * details.
+   */
+  public interface AuthorizedViewType {}
```
kolea2, line 121: picking up convo from last comment, here are the docs for both `@InternalApi` and `@InternalExstensionOnly`:

https://cloud.google.com/java/docs/reference/api-common/latest/com.google.api.core.InternalExtensionOnly
https://cloud.google.com/java/docs/reference/api-common/latest/com.google.api.core.InternalApi

I'll defer to you on which is more suitable for this interface!

trollyxia: I marked it as `@InternalApi`. Thanks for the info!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
```diff
@@ -0,0 +1,123 @@
...
+  public static AuthorizedView fromProto(
+      @Nonnull com.google.bigtable.admin.v2.AuthorizedView proto) {
+    return new AuthorizedView(proto);
+  }
+
+  private AuthorizedView(@Nonnull com.google.bigtable.admin.v2.AuthorizedView proto) {
+    Preconditions.checkNotNull(proto);
+    Preconditions.checkArgument(!proto.getName().isEmpty(), "AuthorizedView must have a name");
+    Preconditions.checkArgument(
+        proto.hasSubsetView(), "AuthorizedView must have a subset_view field");
```
mutianf, line 54: should we have this check in the ctor? looks like when we're getting the view type (line 82) we throw an exception anyways

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
mutianf: I think it should be `@InternalExtensionOnly`? We don't want users to create random subclass, but we want this to be public so users can create a view type?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableAuthorizedViewIT.java
```diff
@@ -0,0 +1,257 @@
...
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class BigtableAuthorizedViewIT {
```
mutianf, line 54: Can we also include a test where FamilySubsets is tested? (with both qualifier and qualifierPrefix)

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/UpdateAuthorizedViewRequestTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+import com.google.cloud.bigtable.admin.v2.internal.NameUtil;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.FieldMask;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class UpdateAuthorizedViewRequestTest {
```
mutianf, line 29: Can we also add a test to update the type to FamilySubsets?

mutianf (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
trollyxia: It is possible that the user constructs an AuthorizedView object but never calls its `getAuthorizedViewType() ` method to examine what's inside it.

So I would prefer keeping the check there.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AuthorizedView.java
trollyxia: Done

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableAuthorizedViewIT.java
trollyxia: Done.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/models/UpdateAuthorizedViewRequestTest.java
trollyxia: Done.

mutianf (approved):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/UpdateAuthorizedViewRequest.java
```diff
@@ -0,0 +1,151 @@
...
+          .setSubsetView(((SubsetView) authorizedViewType).toProto());
+      updateFieldMask(com.google.bigtable.admin.v2.AuthorizedView.SUBSET_VIEW_FIELD_NUMBER);
+    } else {
+      throw new IllegalArgumentException("Unknown authorizedViewType: " + authorizedViewType);
+    }
+
+    return this;
+  }
+
+  /** Configures if safety warnings should be disabled. */
```
igorbernstein2, line 105: nit. please expand what safety means here

igorbernstein2 (changes requested):
Looks pretty good. I think the only thing that might be missing is to update the cleanupStale rule:

https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/TestEnvRule.java#L182

trollyxia:
> Looks pretty good. I think the only thing that might be missing is to update the cleanupStale rule:
> 
> https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/test_helpers/env/TestEnvRule.java#L182

I don't think we need to update that since when a table is deleted, all the authorized views inside that table are also cascading deleted.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/UpdateAuthorizedViewRequest.java
trollyxia: Done.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/FamilySubsets.java
```diff
@@ -0,0 +1,118 @@
...
+  private FamilySubsets(@Nonnull com.google.bigtable.admin.v2.AuthorizedView.FamilySubsets proto) {
+    this.builder = proto.toBuilder();
+  }
+
+  private FamilySubsets() {
+    this.builder = com.google.bigtable.admin.v2.AuthorizedView.FamilySubsets.newBuilder();
+  }
+
+  /** Gets the list of column qualifiers included in this authorized view. */
+  public ImmutableList<ByteString> getQualifiers() {
```
mutianf, line 52: Change the API to return List<ByteString> cause we don't want to leak guava dependencies:
```
public List<ByteString> getQualifiers() {
   return ImmutableList.copyOf(...);
}
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/FamilySubsets.java
```diff
@@ -0,0 +1,118 @@
...
+    return ImmutableList.copyOf(this.builder.getQualifiersList());
+  }
+
+  /** Gets the list of column qualifier prefixes included in this authorized view. */
+  public ImmutableList<ByteString> getQualifierPrefixes() {
```
mutianf, line 57: Same here.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/SubsetView.java
```diff
@@ -0,0 +1,131 @@
...
+  private SubsetView(@Nonnull com.google.bigtable.admin.v2.AuthorizedView.SubsetView proto) {
+    this.builder = proto.toBuilder();
+  }
+
+  private SubsetView() {
+    this.builder = com.google.bigtable.admin.v2.AuthorizedView.SubsetView.newBuilder();
+  }
+
+  /** Gets the row prefixes to be included in this subset view. */
+  public ImmutableList<ByteString> getRowPrefixes() {
```
mutianf, line 61: same here

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/SubsetView.java
```diff
@@ -0,0 +1,131 @@
...
+    return ImmutableList.copyOf(this.builder.getRowPrefixesList());
+  }
+
+  /** Gets the map from familyId to familySubsets in this subset view. */
+  public ImmutableMap<String, FamilySubsets> getFamilySubsets() {
```
mutianf, line 66: same here

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/SubsetView.java
```diff
@@ -0,0 +1,131 @@
...
+  /** Adds a new rowPrefix to the subset view. */
+  public SubsetView addRowPrefix(String rowPrefix) {
+    this.builder.addRowPrefixes(ByteString.copyFromUtf8(rowPrefix));
+    return this;
+  }
+
+  /** Adds a new familyId with its familySubsets to the subset view. */
+  public SubsetView addFamilySubsets(String familyId, FamilySubsets familySubsets) {
+    if (this.builder.containsFamilySubsets(familyId)) {
+      // If the familyId exists, append the familySubsets to the existing value.
```
mutianf, line 90: should the default behavior be replacing instead of merging?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/it/BigtableAuthorizedViewIT.java
```diff
@@ -0,0 +1,272 @@
...
+  @Test
+  public void createAuthorizedViewAndGetAuthorizedViewTest() {
+    String authorizedViewId = prefixGenerator.newPrefix();
+
+    CreateAuthorizedViewRequest request =
+        CreateAuthorizedViewRequest.of(testTable.getId(), authorizedViewId)
+            .setAuthorizedViewType(
+                SubsetView.create()
+                    .addRowPrefix("row#")
+                    .addFamilySubsets("cf1", FamilySubsets.create().addQualifier("qualifier"))
```
mutianf, line 98: once we update the logic to replace, we probably also need to update this test

mutianf (approved):

igorbernstein2 (approved):

PR #2192, opened 2024-04-02
samples: Add AuthorizedView examples

trollyxia:
Change-Id: I77bb799eb448275928bcf0b562c0d4e06870c96d

### DIFF
--- a/samples/snippets/src/main/java/com/example/bigtable/AuthorizedViewExample.java
+++ b/samples/snippets/src/main/java/com/example/bigtable/AuthorizedViewExample.java
@@ -1,0 +1,323 @@
+/*
+ * Copyright 2024 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.example.bigtable;
+
+import static com.google.cloud.bigtable.data.v2.models.Filters.FILTERS;
+
+import com.google.api.gax.rpc.NotFoundException;
+import com.google.api.gax.rpc.ServerStream;
+import com.google.cloud.bigtable.admin.v2.BigtableTableAdminClient;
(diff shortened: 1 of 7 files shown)


samples/snippets/src/main/java/com/example/bigtable/AuthorizedViewExample.java
```diff
@@ -0,0 +1,323 @@
...
+        String greeting = "Hello " + names[i] + "!";
+        RowMutation rowMutation =
+            RowMutation.create(AuthorizedViewId.of(tableId, authorizedViewId), ROW_KEY_PREFIX + i)
+                .setCell(COLUMN_FAMILY, COLUMN_QUALIFIER_NAME, names[i])
+                .setCell(COLUMN_FAMILY, COLUMN_QUALIFIER_GREETING, greeting);
+        dataClient.mutateRow(rowMutation);
+        System.out.println(greeting);
+      }
+    } catch (NotFoundException e) {
+      System.err.println("Failed to write to non-existent authorized view: " + e.getMessage());
```
mutianf, line 267: Maybe update this to "Failed to write to non-existent table or authorized view: " .. cause I think if the table doesn't exist, it would also be NotFound?

samples/snippets/src/main/java/com/example/bigtable/AuthorizedViewExample.java
```diff
@@ -0,0 +1,323 @@
...
+    }
```
mutianf, line 268: Do we also need to catch permission denied? If the writer doesn't have write permission to the view? Same for other examples.

samples/snippets/src/main/java/com/example/bigtable/AuthorizedViewExample.java
trollyxia: Hmmm but here we are targeting the authorized view only. Therefore, in the AFE we would perform spanner lookups for the authorized view. So if the parent table doesn't exist, the NotFound error returned from the AFE is because of the missing authorized view instead of the missing table.
 
It's similar to when we are trying to write to a table and got an NotFound error, we wouldn't print the error msg as "Failed to write to non-existent table or instance"?

samples/snippets/src/main/java/com/example/bigtable/AuthorizedViewExample.java
trollyxia: Done.

The other two read-only examples wouldn't receive permission denied error for reads outside of the authorized view. The server would simply return empty rows.

samples/snippets/src/main/java/com/example/bigtable/AuthorizedViewExample.java
mutianf: Ok, make sense :)

PR #2033, opened 2023-12-18
fix: remove grpclb

mohanli-ml:
Remove attempt DirectPath gRPCLB by default in CBT Java because:
1. gRPCLB is being deprecated, and CBT has removed all the gRPCLB traffic back to CloudPath;
2. Attempt gRPCLB by default has known DNS amplifying issue, https://github.com/googleapis/google-cloud-go/issues/7812.

PR #2263, opened 2024-06-24
test: ensure that directpath is enabled for direct path tests

igorbernstein2:
This should unblock #2033 which disables directpath by default


kongweihan (approved):
LGTM

PR #2287, opened 2024-07-23
feat: Add support for new functions

ron-gal:
- [x] Rollback plan is reviewed and LGTMed
- [x] All new data plane features have a completed end to end testing plan


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/Type.java
```diff
@@ -87,13 +92,48 @@ public static Aggregate int64Sum() {
...
+    return max(bigEndianInt64());
+  }
+
+  /** Creates an Aggregate type with a MAX aggregator and specified input type. */
+  public static <Input extends Type & MaxAggregateInput> Aggregate max(Input inputType) {
+    return Aggregate.create(inputType, Aggregate.Aggregator.Max.create());
+  }
+
+  /** Creates an Aggregate type with a HLL aggregator and Int64 input type. */
+  public static Aggregate rawBytesHll() {
```
steveniemitz, line 125: shouldn't this be called int64Hll?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/Type.java
```diff
@@ -87,13 +92,48 @@ public static Aggregate int64Sum() {
...
+
+  /** Creates an Aggregate type with a HLL aggregator and specified input type. */
+  public static <Input extends Type & HllAggregateInput> Aggregate hll(Input inputType) {
+    return Aggregate.create(inputType, Aggregate.Aggregator.Hll.create());
+  }
+
   /** Represents a string of bytes with a specific encoding. */
   @AutoValue
-  public abstract static class Bytes extends Type {
+  public abstract static class Bytes extends Type implements HllAggregateInput {
```
steveniemitz, line 136: same, we only support int64 inputs right now so we shouldn't have this here.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/Type.java
ron-gal: Whoops! D= fixed!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/Type.java
ron-gal: fixed

steveniemitz (approved):

PR #2312, opened 2024-08-15
feat: allow non default service account in DirectPath

meeral-k:
Copying over from: https://github.com/googleapis/java-bigtable/pull/2012

Bigtable already sets the flag to allow non-default SA in DirectPath on the service side, we also need to set the option on the client side to make this feature work.
Related Spanner PR: https://github.com/googleapis/java-spanner/pull/2635.


meeral-k:
@mohanli-ml fyi

igorbernstein2 (approved):

PR #2317, opened 2024-08-19
chore: secure hermetic_library_generation workflow

diegomarquezp:
Thanks to @diogoteles08 for the inspection on our repos. This PR inlines environment variables to avoid overriding script injections.

### DIFF
--- a/.github/workflows/hermetic_library_generation.yaml
+++ b/.github/workflows/hermetic_library_generation.yaml
@@ -20,24 +20,10 @@ on:
 env:
   HEAD_REF: ${{ github.head_ref }}
   REPO_FULL_NAME: ${{ github.event.pull_request.head.repo.full_name }}
+  GITHUB_REPOSITORY: ${{ github.repository }}
 
 jobs:
-  # This step is meant to inline the the repo full name env var in the `jobs.if` entry
-  # of the next step "library_generation".
-  # Unfortunately, we can't directly reference `env` in jobs.if
-  # (see https://github.com/actions/runner/issues/2372)
-  # The chosen workaround is to declare an intermediate step that will output the env var
-  prepare-repo-full-name:
-    name: Prepare environment
-    runs-on: ubuntu-latest
-    outputs:
-      REPO_FULL_NAME: ${{ env.REPO_FULL_NAME }}
-    steps:
-      - run: exit 0 # no op - we just need to declare the env var as output
   library_generation:
-    # skip pull requests coming from a forked repository
-    if: needs.prepare-repo-full-name.outputs.REPO_FULL_NAME == github.repository
-    needs: [prepare-repo-full-name]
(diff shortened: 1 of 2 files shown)


.github/workflows/hermetic_library_generation.yaml
```diff
@@ -19,11 +19,25 @@ on:
...
+  # Unfortunately, we can't directly reference `env` in jobs.if
+  # (see https://github.com/actions/runner/issues/2372)
+  # The chosen workaround is to declare an intermediate step that will output the env var
+  prepare-repo-full-name:
+    name: Prepare environment
+    runs-on: ubuntu-latest
+    outputs:
+      REPO_FULL_NAME: ${{ env.REPO_FULL_NAME }}
+    steps:
+      - run: echo "" >> /dev/null # no op - we just need to declare the env var as output
```
blakeli0, line 36: Is this empty step required?

diegomarquezp: Unfortunately, yes
image

However, we can make it even simpler (`exit 0`)

.github/workflows/hermetic_library_generation.yaml
```diff
@@ -19,11 +19,25 @@ on:
...
+    runs-on: ubuntu-latest
+    outputs:
+      REPO_FULL_NAME: ${{ env.REPO_FULL_NAME }}
+    steps:
+      - run: exit 0 # no op - we just need to declare the env var as output
   library_generation:
-    # skip pull requests come from a forked repository
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    # skip pull requests coming from a forked repository
+    if: needs.prepare-repo-full-name.outputs.REPO_FULL_NAME == github.repository
```
blakeli0, line 39: What if we move this if condition to the `run` section? Did we consider the approach before @JoeWang1127 ?

diegomarquezp: I modified it to just have an `if` inside `run`. I'll try a fork PR into this branch

diegomarquezp: I tried raising [a PR from a fork](https://github.com/googleapis/java-bigtable/pull/2325)
image

I tried setting the missing token in my fork without luck.
However I could confirm that the generation was properly triggered in this PR because it doesn't come from a fork.
https://github.com/googleapis/java-bigtable/actions/runs/10562926017/job/29261969452?pr=2317

JoeWang1127:
Could you create a forked repo and test whether the workflow will be skipped?

diegomarquezp:
> Could you create a forked repo and test whether the workflow will be skipped?

@JoeWang1127 we tried this in https://github.com/googleapis/java-bigtable/pull/2317#discussion_r1731458621. I think we still confirmed it works the other way around: non-forked repos actually get triggered.

PR #2341, opened 2024-09-16
feat: add support for Row Affinity app profiles

djyau:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AppProfile.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AppProfile.java
@@ -69,6 +69,10 @@ public final class AppProfile {
   @SuppressWarnings("WeakerAccess")
   public RoutingPolicy getPolicy() {
     if (proto.hasMultiClusterRoutingUseAny()) {
+      if (proto.getMultiClusterRoutingUseAny().hasRowAffinity()) {
+        return MultiClusterRoutingPolicy.ofWithRowAffinity(
+          ImmutableSet.copyOf(proto.getMultiClusterRoutingUseAny().getClusterIdsList()));
+      }
       return MultiClusterRoutingPolicy.of(
           ImmutableSet.copyOf(proto.getMultiClusterRoutingUseAny().getClusterIdsList()));
     } else if (proto.hasSingleClusterRouting()) {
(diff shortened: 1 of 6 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/AppProfile.java
```diff
@@ -267,6 +271,32 @@ public static MultiClusterRoutingPolicy of(Set<String> clusterIds) {
           MultiClusterRoutingUseAny.newBuilder().addAllClusterIds(clusterIds).build());
     }
 
+    /** Creates a new instance of {@link MultiClusterRoutingPolicy}. */
+    public static MultiClusterRoutingPolicy ofWithRowAffinity() {
```
igorbernstein2, line 275: I think withRowAffinity is a bit less akward

igorbernstein2: Also can you describe what the difference between this variant and withRowAffinity(String... clusterIds)

igorbernstein2: Also is it possible to update an app profile to change cluster ids? or to change the routing away from row affinity?

igorbernstein2: Also do you need to update the mask?

djyau: > Also can you describe what the difference between this variant and withRowAffinity(String... clusterIds)

Yes - Row affinity, when enabled, routes single row key requests deterministically to the same cluster each time to improve read-your-writes consistency.

When row affinity is enabled with a cluster group (with cluster IDs specified), then when the user adds a new cluster to their instance, it does not affect which clusters each row key routes to. It is a way to protect read-your-writes consistency from being impacted negatively from new clusters as well as from attempts to delete clusters (since you cannot delete a cluster if its in the cluster group). Without the cluster group, when a user adds a new cluster, row keys will undergo a shuffling of which cluster they route to, which temporarily breaks read-your-writes consistency.

djyau: > Also do you need to update the mask?

I don't think so, but correct me if I am wrong: row affinity is a subfield/submessage of MultiClusterRoutingUseAny (routing policy), and we already have existing methods for updating the routing policy with the appropriate mask: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/UpdateAppProfileRequest.java#L106-L125

> Also is it possible to update an app profile to change cluster ids? or to change the routing away from row affinity?

It should be possible to do both of these. The user should call UpdateAppProfileRequest::SetRoutingPolicy and supply a routing policy with new cluster ids or with row affinity absent

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientTests.java
```diff
@@ -1034,6 +1034,135 @@ public void testCreateAppProfileAddPriority() {
...
+
+    Mockito.when(mockCreateAppProfileCallable.futureCall(expectedRequest))
+            .thenReturn(ApiFutures.immediateFuture(expectedResponse));
+
+    // Execute
+    AppProfile actualResult =
+            adminClient.createAppProfile(
+                    CreateAppProfileRequest.of(INSTANCE_ID, APP_PROFILE_ID)
+                            .setDescription("my description")
+                            .setRoutingPolicy(MultiClusterRoutingPolicy.withRowAffinity("cluster-id-1", "cluster-id-2")));
```
mutianf, line 1160: Is this the same as the test above?

djyau: It is similar. The one above adds the cluster_ids one at a time via "addClusterIds". This test adds all the cluster_ids in one list via "addAllClusterIds"

mutianf: I think that's testing the generated proto instead of testing our wrapper class. We can probably just keep one of the 2 test.

djyau: Okie - removed this one.

mutianf (approved):

PR #2346, opened 2024-09-19
fix: fix client blocking latency

mutianf:

### DIFF
--- a/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/CompositeTracerTest.java
+++ b/google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/CompositeTracerTest.java
@@ -258,4 +258,11 @@ public class CompositeTracerTest {
     verify(child3, times(1)).grpcChannelQueuedLatencies(5L);
     verify(child4, times(1)).grpcChannelQueuedLatencies(5L);
   }
+
+  @Test
+  public void testGrpcMessageSent() {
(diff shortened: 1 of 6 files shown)


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/CompositeTracerTest.java
```diff
@@ -258,4 +258,11 @@ public void testRequestBlockedOnChannel() {
     verify(child3, times(1)).grpcChannelQueuedLatencies(5L);
     verify(child4, times(1)).grpcChannelQueuedLatencies(5L);
   }
+
+  @Test
+  public void testGrpcMessageSent() {
+    compositeTracer.grpcMessageSent();
+    verify(child3, times(1)).grpcMessageSent();
+    verify(child4, times(1)).grpcMessageSent();
+  }
```
igorbernstein2, line 267: please add a test that actually tests for the old broken behavior. Specifically to ensure that client blocking latencies take delays into account. I can think of 2 ways to this:

1. Disable ping & warm and try to delay the client connection. One way I can think of is to use a ChannelConfigurator that injects a ProxyDetector that injects a sleep into proxyFor method before returning null
2. Open more that 100 concurrent rpcs to an in process server. Sleep for X time then resolve all of the rpcs and ensure that the last one recorded the proper client blocking latency

igorbernstein2 (approved):

PR #2360, opened 2024-09-27
Add test proxy implementation for ExecuteQuery api

jackdingilian:


test-proxy/src/main/proto/test_proxy.proto
```diff
@@ -66,6 +87,17 @@ message CreateClientRequest {
...
+  // Options to allow connecting to backends with channel and/or call
+  // credentials. This is needed internally by Cloud Bigtable's own testing
+  // frameworks.It is not necessary to support these fields for client
+  // conformance testing.
+  //
+  // WARNING: this allows the proxy to connect to a real production
+  // CBT backend with the right options, however, the proxy itself is insecure
+  // so it is not recommended to use it with real credentials or outside testing
+  // contexts.
+  SecurityOptions security_options = 8;
```
mutianf, line 100: Can these changes be in a separate PR? able to connect to production also sounds scary. Is there a way to avoid that?

test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/ResultSetSerializer.java
```diff
@@ -0,0 +1,233 @@
...
+import com.google.cloud.bigtable.data.v2.models.sql.ColumnMetadata;
+import com.google.cloud.bigtable.data.v2.models.sql.ResultSet;
+import com.google.cloud.bigtable.data.v2.models.sql.SqlType;
+import com.google.cloud.bigtable.data.v2.models.sql.StructReader;
+import com.google.protobuf.ByteString;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import org.threeten.bp.Instant;
+
+public class ResultSetSerializer {
```
mutianf, line 40: Are there existing client code for serialization and deserialization? This is a lot of code to maintain and test proxy should just be a thin layer that calls into the client

jackdingilian: There is client code (within ResultSet) that handles deserializing protobuf Value messages to the corresponding java types. I agree this feels too 'heavy' for the testproxy but I think with the introduction of types it is somewhat unavoidable. 

We want to validate that each client is converting to the relevant types correctly, to do so we need to do some form of roundtripping back to a shared format we can validate against. We reuse Value from the api here. I think no matter how we do it we need to do some conversion from java type back to some protobuf format.

jackdingilian: For existing java type -> Value serialization theres some existing code that does a subset of what we're doing here but it's not sufficient to cover what we need and Im not sure either is a great fit for replacing this.

1) https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Value.java#L112 - This serializes some Values to proto but is only used for the types supported by aggregates right now. We would have no use for the additional SQL types there and we would need to have the aggregates apis reject those value types (https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/RowMutation.java#L231)

2) https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/sql/Statement.java#L194

Statement does this for the types of query params we support but it doesn't support all the types we need (no map or struct) and sets the Value.type field which we don't expect here (though I guess it would be ok)

test-proxy/src/main/proto/test_proxy.proto
jackdingilian: Talked to Patrick and removed this for now. We do want to use this for some internal tests, so we will put it in a separate PR

mutianf (approved):

PR #2374, opened 2024-10-08
fix: allow factory to export to different projects

mutianf:
This fix removed the check on Bigtable project id and gets the BigtableTable resource project id directly from metrics attribute. BigtableDataClientFactory can create one client for multiple projects. Removing the check allows people using BigtableDataClientFactory to export to different projects.


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -126,8 +126,7 @@ public final class BigtableCloudMonitoringExporter implements MetricExporter {
           .collect(ImmutableList.toImmutableList());
 
   public static BigtableCloudMonitoringExporter create(
-      String projectId, @Nullable Credentials credentials, @Nullable String endpoint)
-      throws IOException {
+      @Nullable Credentials credentials, @Nullable String endpoint) throws IOException {
```
igorbernstein2, line 129: Im a lil concerned about the remaining args here - like the problematic projectId, they are used to construct a singleton, but not consulted when getting existing instances of the singleton. What happens if someone constructs 2 clients with a different set of creds?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -209,15 +203,7 @@ private CompletableResultCode exportBigtableResourceMetrics(Collection<MetricDat
...
-    // Verifies metrics project id are the same as the bigtable project id set on this client
-    if (!bigtableMetricData.stream()
-        .flatMap(metricData -> metricData.getData().getPoints().stream())
-        .allMatch(pd -> bigtableProjectId.equals(BigtableExporterUtils.getProjectId(pd)))) {
-      logger.log(Level.WARNING, "Metric data has different a projectId. Skip exporting.");
-      return CompletableResultCode.ofFailure();
-    }
-
-    List<TimeSeries> bigtableTimeSeries;
+    Map<String, List<TimeSeries>> bigtableTimeSeries;
```
igorbernstein2, line 206: please add a comment what the key is

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -110,17 +111,22 @@ static String getProjectId(PointData pointData) {
     return pointData.getAttributes().get(BIGTABLE_PROJECT_ID_KEY);
   }
 
-  static List<TimeSeries> convertToBigtableTimeSeries(List<MetricData> collection, String taskId) {
-    List<TimeSeries> allTimeSeries = new ArrayList<>();
+  static Map<String, List<TimeSeries>> convertToBigtableTimeSeries(
```
igorbernstein2, line 114: docs for what the key is here

igorbernstein2 (changes requested):
also, this needs a test :)

mutianf:
Note: global otel is not idea. We should refactor how the client context is created so otel doesn't need to be global.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableClientContext.java
```diff
@@ -0,0 +1,170 @@
...
+
+    clientContext = ClientContext.create(builder.build());
+
+    if (errorCountPerConnectionMetricTracker != null) {
+      errorCountPerConnectionMetricTracker.startConnectionErrorCountTracker(
+          clientContext.getExecutor());
+    }
+  }
+
+  public OpenTelemetry createOpenTelemetry() {
```
igorbernstein2, line 129: this doesnt look like it creates anything :)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableClientContext.java
```diff
@@ -0,0 +1,170 @@
...
+      this.openTelemetry =
+          createOpenTelemetry(
+              settings.getMetricsProvider(), credentials, settings.getMetricsEndpoint());
+    } catch (Throwable t) {
+      logger.log(Level.WARNING, "Failed to get OTEL, will skip exporting client side metrics", t);
+    }
+    ErrorCountPerConnectionMetricTracker errorCountPerConnectionMetricTracker;
+    // Skip setting up ErrorCountPerConnectionMetricTracker if openTelemetry is null
+    if (openTelemetry != null && transportProvider != null) {
+      errorCountPerConnectionMetricTracker =
```
igorbernstein2, line 82: is it possible to factor this out to a helper method?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClientFactory.java
```diff
@@ -132,10 +113,11 @@ public BigtableDataClient createDefault() {
     try {
       ClientContext clientContext =
           sharedClientContext
+              .getClientContext()
               .toBuilder()
               .setTracerFactory(
```
igorbernstein2, line 118: does the TracerFactory need to be overwritten for the default case?

PR #2382, opened 2024-10-17
fix: fix first response latencies

mutianf:


google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracerTest.java
```diff
@@ -327,6 +329,51 @@ public void testReadRowsOperationLatenciesOnAuthorizedView() {
...
+            new ResponseObserver<Row>() {
+              @Override
+              public void onStart(StreamController controller) {}
+
+              @Override
+              public void onResponse(Row response) {
+                if (firstResponseTimer.isRunning()) {
+                  firstResponseTimer.stop();
+                }
+                try {
```
igorbernstein2, line 347: This is pretty fragile - you are assuming that operation timer will not stop on onComplete while the caller is blocking onResponse.

I think it would be much safer to send 2 responses with a delay

mutianf: Server is sending back 2 responses so I think this is fine? Added a comment

igorbernstein2 (approved):

PR #2394, opened 2024-11-01
chore: refactor unary callable chains to avoid code duplication

igorbernstein2:
This is primarily a code clean up change that extracts all of the unary chain building logic into a helper method and removes a significant amount of code duplication.


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -1404,6 +1324,57 @@ private <RequestT, ResponseT> UnaryCallable<RequestT, ResponseT> createUserFacin
...
+                .build(),
+            callSettings.getRetryableCodes());
+
+    UnaryCallable<BaseReqT, BaseRespT> withStatsHeaders = new StatsHeadersUnaryCallable<>(base);
+
+    UnaryCallable<BaseReqT, BaseRespT> withBigtableTracer =
+        new BigtableTracerUnaryCallable<>(withStatsHeaders);
+
+    UnaryCallable<BaseReqT, BaseRespT> retrying =
+        withRetries(withBigtableTracer, settings.mutateRowSettings());
```
mutianf, line 1356: should be withRetries(withBigtableTracer, callSettings)?

mutianf (changes requested):
Can we add the unary tests to EnhancedBigtableStubTest?

mutianf (approved):

PR #2403, opened 2024-11-04
chore: remodel unary callables as server streaming callables with an adapter at the end

igorbernstein2:
This allows us to decouple payload and trailer handling.


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableUnaryOperationCallable.java
```diff
@@ -0,0 +1,189 @@
...
+      this.tracer = Preconditions.checkNotNull(tracer, "tracer can't be null");
+      this.allowNoResponse = allowNoResponse;
+      this.responseReceived = false;
+    }
+
+    @Override
+    public void onStart(StreamController controller) {
+      this.controller = controller;
+      controller.disableAutoInboundFlowControl();
+      // Request 2 to detect protocol bugs
```
mutianf, line 106: why? is this necessary?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableUnaryOperationCallable.java
```diff
@@ -0,0 +1,189 @@
...
+        if (setException(e)) {
+          tracer.operationFailed(e);
+          return;
+        }
+      }
+
+      // check cancellation race
+      if (isCancelled()) {
+        tracer.operationCancelled();
+      }
```
mutianf, line 186: Should we also do tracer.operationFailed here?  if all of (set(Response)) or setException(e) and isCancelled returned false we'll end up not calling operation completion in the tracer? I'm not sure when we'll hit this case though. Same question for onError. Do we trust that onComplete or onError is already called by another thread?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -1361,6 +1407,50 @@ public ApiFuture<RespT> futureCall(ReqT reqT, ApiCallContext apiCallContext) {
...
+                .setMethodDescriptor(methodDescriptor)
+                .setParamsExtractor(headerParamsFn)
+                .build(),
+            callSettings.getRetryableCodes());
+
+    base = new StatsHeadersServerStreamingCallable<>(base);
+
+    base = new BigtableTracerStreamingCallable<>(base);
+
+    base = withRetries(base, convertUnaryToServerStreamingSettings(settings.mutateRowSettings()));
```
mutianf, line 1429: mutateRowSettings -> callSettings :)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -1361,6 +1407,50 @@ public ApiFuture<RespT> futureCall(ReqT reqT, ApiCallContext apiCallContext) {
...
+  }
+
+  private static <ReqT, RespT>
+      ServerStreamingCallSettings<ReqT, RespT> convertUnaryToServerStreamingSettings(
+          UnaryCallSettings<?, ?> unarySettings) {
+    return ServerStreamingCallSettings.<ReqT, RespT>newBuilder()
+        .setResumptionStrategy(new SimpleStreamResumptionStrategy<>())
+        .setRetryableCodes(unarySettings.getRetryableCodes())
+        .setRetrySettings(unarySettings.getRetrySettings())
+        .setIdleTimeoutDuration(Duration.ZERO)
```
mutianf, line 1449: In readRow, idle timeout is total timeout. Should we change that to be zero too?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/TransformingServerStreamingCallable.java
```diff
@@ -0,0 +1,72 @@
...
+package com.google.cloud.bigtable.data.v2.stub;
+
+import com.google.api.gax.rpc.ApiCallContext;
+import com.google.api.gax.rpc.ResponseObserver;
+import com.google.api.gax.rpc.ServerStreamingCallable;
+import com.google.api.gax.rpc.StreamController;
+import java.util.function.Function;
+
+/** Callable to help crossing api boundary lines between models and protos */
+class TransformingServerStreamingCallable<OuterReqT, OuterRespT, InnerReqT, InnerRespT>
```
mutianf, line 25: Do we need this class? Or we can merge it to BigtableUnaryOperationCallable? Cause all we're doing here is transform the request and response, which can be done in BigtableUnaryOperationCallable

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableUnaryOperationCallable.java
```diff
@@ -0,0 +1,189 @@
...
+ * UnaryCallable}. It is intended to be the outermost callable of a chain.
+ *
+ * <p>Responsibilities:
+ *
+ * <ul>
+ *   <li>Operation level metrics
+ *   <li>Configuring the default call context
+ *   <li>Converting the result to a future
+ */
+class BigtableUnaryOperationCallable<ReqT, RespT> extends UnaryCallable<ReqT, RespT> {
```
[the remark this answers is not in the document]
igorbernstein2: I'm trying to make the names semantic, the eventual goal is to have 2 endcaps for all of the callable chains
- BigtableUnaryOperationCallable
- BigtableSearverStreamingOperationCallable

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableUnaryOperationCallable.java
```diff
@@ -0,0 +1,189 @@
...
+            apiCallContext.getTracer(), spanName, ApiTracerFactory.OperationType.Unary);
+
+    apiCallContext = apiCallContext.withTracer(apiTracer);
+
+    UnaryFuture f = new UnaryFuture(apiTracer, allowNoResponse);
+    inner.call(req, f, apiCallContext);
+    return f;
+  }
+
+  class UnaryFuture extends AbstractApiFuture<RespT> implements ResponseObserver<RespT> {
```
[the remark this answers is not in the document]
igorbernstein2: it cant because it already extends AbstractApiFuture

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableUnaryOperationCallable.java
igorbernstein2: to preserve the current behavior

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableUnaryOperationCallable.java
igorbernstein2: no, because that will double call the tracer

if all of those methods returned false it means we've already notified the tracer.  

At the beginning of the method we can be in one of 2 states:
1. unresolved (happy path)
2. error (onResponse was called multiple times)
3. cancelled (user cancelled the future)

case#1 is straightfoward and is handled by the set on L168
case#2 means that onResponse setException() and paired it with a tracer notification, so we dont want to notify the tracer again

case#3 would make set & stException return false and we explicitly detect it on 184

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
igorbernstein2: bah, thanks for catching it

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
igorbernstein2: not sure what you mean?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/TransformingServerStreamingCallable.java
igorbernstein2: We will need it in future refactoring where the callable chain does a transform in the middle of the chain

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
igorbernstein2: talked offline, fixed

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableUnaryOperationCallable.java
igorbernstein2: https://github.com/grpc/grpc-java/blob/dae078c0a636fd4b71ca7d20c333dd00b0b0714a/stub/src/main/java/io/grpc/stub/ClientCalls.java#L544

mutianf (approved):

PR #2404, opened 2024-11-05
feat: add an experimental feature to skip waiting for trailers for unary ops

igorbernstein2:
This is off by default and can be enabled using an environment variable.
When enabled, BigtableUnaryOperationCallable will resolve the user visible future immediately when a response is available and will tell metrics to freeze all timers. Metrics will still wait for the trailers in the background for necessary metadata to publish the frozen timer values.

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java
@@ -299,9 +299,7 @@ class BuiltinMetricsTracer extends BigtableTracer {
     if (!opFinished.compareAndSet(false, true)) {
       return;
     }
-    if (operationTimer.isRunning()) {
-      operationTimer.stop();
-    }
+    long operationLatencyNano = operationTimer.elapsed(TimeUnit.NANOSECONDS);
 
     boolean isStreaming = operationType == OperationType.ServerStreaming;
     String statusStr = Util.extractStatus(status);
@@ -319,8 +317,6 @@ class BuiltinMetricsTracer extends BigtableTracer {
             .put(STREAMING_KEY, isStreaming)
             .put(STATUS_KEY, statusStr)
             .build();
-
-    long operationLatencyNano = operationTimer.elapsed(TimeUnit.NANOSECONDS);
 
     // Only record when retry count is greater than 0 so the retry
     // graph will be less confusing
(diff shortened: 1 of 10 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java
```diff
@@ -272,10 +292,16 @@ public void disableFlowControl() {
...
   private void recordOperationCompletion(@Nullable Throwable status) {
+    if (operationFinishedEarly.get()) {
+      status = null; // force an ok
+    }
+
     if (!opFinished.compareAndSet(false, true)) {
       return;
     }
-    operationTimer.stop();
+    if (operationTimer.isRunning()) {
```
mutianf, line 302: I think it's unnecessary to check if the timer is still running and stop it, we can just do  `long operationLatencyNano = operationTimer.elapsed(TimeUnit.NANOSECONDS);`.  (sorry about my redundant code)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracer.java
```diff
@@ -316,6 +342,9 @@ private void recordOperationCompletion(@Nullable Throwable status) {
   }
 
   private void recordAttemptCompletion(@Nullable Throwable status) {
+    if (operationFinishedEarly.get()) {
```
mutianf, line 345: We probably want to move this to line 363 

```
    if (operationFinishedEarly.get()) {
      // force an ok status
       status = null;
    } else if (status instanceof ServerStreamingAttemptException) {
      status = status.getCause();
    }
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/MetricsTracer.java
```diff
@@ -103,7 +109,11 @@ private void recordOperationCompletion(@Nullable Throwable throwable) {
     if (!opFinished.compareAndSet(false, true)) {
       return;
     }
-    operationTimer.stop();
+
+    // Mightve stopped in operationFinishEarly()
+    if (operationTimer.isRunning()) {
```
mutianf, line 114: Same as BuiltinMetricsTracer, I don't think we need this check?

mutianf (approved):

PR #2435, opened 2024-12-02
fix: send priming requests on the channel directly

### LINKED ISSUE #2371: Bigtable: Client side metrics opt-out is not honored in BigtableChannelPrimer
**Issue**
Since `2.38.0`, client-side metrics have been enabled by default. 
The docs [here](https://cloud.google.com/bigtable/docs/client-side-metrics-setup#use-bigtable-api) state that one needs to pass a `NoopMetricsProvider.INSTANCE` in `BigtableDataSettings` to opt out of the metrics.
However client side metrics are anyways enabled during channel priming as `BigtableChannelPrimer` creates its own `EnhancedBigtableStubSettings`.

**Impact**
Local testing with Bigtable test container has been impacted by the timeout of 1 minute used in `BigtableCloudMonitoringExporter` as the exporter is either unable to reach or authenticate with the cloud monitoring endpoint.
Production instances are emitting metrics even if an opt-out was added.

**Known workaround**
For tests, using the deprecated configuration to [disable channel refreshes](https://cloud.google.com/java/docs/reference/google-cloud-bigtable/latest/com.google.cloud.bigtable.data.v2.BigtableDataSettings#com_google_cloud_bigtable_data_v2_BigtableDataSettings_isRefreshingChannel__) prevents creation of `BigtableChannelPrimer`, bypassing the issue. 
This option, however, is due to be removed in future.

**Bigtable library version**
2.44.0 but all versions since 2.38.0 should be affected

mutianf:
Send priming requests on the channel instead of using the stub. This means that we'll not collect metrics on ping and warm requests.

Fixes #2371 ☕️

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
@@ -50,11 +50,11 @@ import java.util.logging.Logger;
 class BigtableChannelPrimer implements ChannelPrimer {
   private static Logger LOG = Logger.getLogger(BigtableChannelPrimer.class.toString());
 
-  private final Metadata.Key<String> requestParams =
+  static final Metadata.Key<String> REQUEST_PARAMS =
       Metadata.Key.of("x-goog-request-params", Metadata.ASCII_STRING_MARSHALLER);
   private final PingAndWarmRequest request;
-  private final Metadata metadata = new Metadata();
   private final CallCredentials callCredentials;
+  private final Map<String, String> headers;
 
   static BigtableChannelPrimer create(
       String projectId,
@@ -83,20 +83,7 @@ class BigtableChannelPrimer implements ChannelPrimer {
             .setAppProfileId(appProfileId)
             .build();
 
-    Metadata metadata = new Metadata();
-    headers.forEach(
-        (k, v) -> metadata.put(Metadata.Key.of(k, Metadata.ASCII_STRING_MARSHALLER), v));
-
-    try {
-      metadata.put(
-          requestParams,
-          String.format(
-              "name=%s&app_profile_id=%s",
-              URLEncoder.encode(request.getName(), "UTF-8"),
-              URLEncoder.encode(request.getAppProfileId(), "UTF-8")));
-    } catch (UnsupportedEncodingException e) {
-      LOG.log(Level.WARNING, "Failed to encode request params", e);
-    }
+    this.headers = headers;
   }
 
   @Override
@@ -104,8 +91,7 @@ class BigtableChannelPrimer implements ChannelPrimer {
     try {
       primeChannelUnsafe(managedChannel);
     } catch (IOException | RuntimeException e) {
-      LOG.warning(
-          String.format("Unexpected error while trying to prime a channel: %s", e.getMessage()));
+      LOG.log(Level.WARNING, "Unexpected error while trying to prime a channel", e);
     }
   }
 
@@ -142,7 +128,7 @@ class BigtableChannelPrimer implements ChannelPrimer {
               }
             }
           },
-          metadata);
+          createMetadata(headers, request));
       clientCall.sendMessage(request);
       clientCall.halfClose();
       clientCall.request(Integer.MAX_VALUE);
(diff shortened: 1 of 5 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
```diff
@@ -41,27 +45,46 @@
...
-      Credentials credentials, String projectId, String instanceId, String appProfileId) {
-    EnhancedBigtableStubSettings.Builder builder =
-        EnhancedBigtableStubSettings.newBuilder()
-            .setProjectId(projectId)
-            .setInstanceId(instanceId)
+  static BigtableChannelPrimer create(EnhancedBigtableStubSettings settings) throws IOException {
+    return new BigtableChannelPrimer(settings);
+  }
+
+  BigtableChannelPrimer(EnhancedBigtableStubSettings settings) throws IOException {
```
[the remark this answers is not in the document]
mutianf: But BigtableClientContext is not created yet when we create BigtableChannelPrimer

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
```diff
@@ -41,27 +50,53 @@
 class BigtableChannelPrimer implements ChannelPrimer {
   private static Logger LOG = Logger.getLogger(BigtableChannelPrimer.class.toString());
 
-  private final EnhancedBigtableStubSettings settingsTemplate;
+  private final Metadata.Key<String> requestParams =
+      Metadata.Key.of("x-goog-request-params", Metadata.ASCII_STRING_MARSHALLER);
+  private final PingAndWarmRequest request;
+  private final Metadata metadata = new Metadata();
```
igorbernstein2, line 56: This is fairly broken in a couple of ways:
1. it doesnt get populated...it gets shadowed in the constructor by a local variable
2. if it wasnt shadowed, then it would break because metadata instances cant be reused across multipple calls

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
```diff
@@ -41,27 +50,53 @@
...
-            .setCredentialsProvider(FixedCredentialsProvider.create(credentials))
-            // Disable refreshing channel here to avoid creating settings in a loop
-            .setRefreshingChannel(false)
-            .setExecutorProvider(
-                InstantiatingExecutorProvider.newBuilder().setExecutorThreadCount(1).build());
+            .build();
 
-    return new BigtableChannelPrimer(builder.build());
-  }
+    Metadata metadata = new Metadata();
```
igorbernstein2, line 86: this looks like it doesnt get used?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
```diff
@@ -78,35 +113,46 @@ private void primeChannelUnsafe(ManagedChannel managedChannel) throws IOExceptio
...
+            @Override
+            public void onClose(Status status, Metadata trailers) {
+              if (status.isOk()) {
+                future.set(response);
+              } else {
+                future.setException(status.asException());
+              }
+            }
+          },
+          metadata);
```
igorbernstein2, line 145: You want to compose a new Metadata isntance per call

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
```diff
@@ -41,72 +50,114 @@
...
+      future.get(1, TimeUnit.MINUTES);
+    } catch (Throwable e) {
+      // TODO: Not sure if we should swallow the error here. We are pre-emptively swapping
+      // channels if the new
+      // channel is bad.
+      LOG.warning(String.format("Failed to prime channel: %s", e));
     }
   }
+
+  private Metadata createMetadata(Map<String, String> headers, PingAndWarmRequest request) {
```
igorbernstein2, line 145: static

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
```diff
@@ -41,72 +50,114 @@
...
+      clientCall.sendMessage(request);
+      clientCall.halfClose();
+      clientCall.request(Integer.MAX_VALUE);
+
+      future.get(1, TimeUnit.MINUTES);
+    } catch (Throwable e) {
+      // TODO: Not sure if we should swallow the error here. We are pre-emptively swapping
+      // channels if the new
+      // channel is bad.
+      LOG.warning(String.format("Failed to prime channel: %s", e));
```
igorbernstein2, line 141: - remove string.format
- add the missing format param:
LOG.warning("Failed to prime channel {}", managedChannel, e);

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableChannelPrimer.java
```diff
@@ -41,72 +50,114 @@
...
-            .toBuilder()
-            .setTransportChannelProvider(
-                FixedTransportChannelProvider.create(GrpcTransportChannel.create(managedChannel)))
-            .build();
+  private void sendPrimeRequests(ManagedChannel managedChannel) {
+    try {
+      ClientCall<PingAndWarmRequest, PingAndWarmResponse> clientCall =
+          managedChannel.newCall(
+              BigtableGrpc.getPingAndWarmMethod(),
+              GrpcCallContext.createDefault()
```
igorbernstein2, line 107: why do you need gax wrappers here? I think you want CallOptions.DEFAULT

PR #2469, opened 2025-01-16
feat: Adding default retry settings for consistency check

sandshar09:


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/BigtableTableAdminStubSettings.java
```diff
@@ -940,6 +945,17 @@ public static class Builder
...
               .build();
       definitions.put("retry_policy_2_params", settings);
+      settings =
+          RetrySettings.newBuilder()
+              .setInitialRetryDelayDuration(Duration.ofMillis(1000L))
+              .setRetryDelayMultiplier(2.0)
+              .setMaxRetryDelayDuration(Duration.ofMillis(60000L))
+              .setInitialRpcTimeoutDuration(Duration.ofMillis(3600000L))
+              .setRpcTimeoutMultiplier(1.0)
+              .setMaxRpcTimeoutDuration(Duration.ofMillis(3600000L))
```
justinuang, line 955: I think this could be set back to 60000L, since it's not a single RPC we're concerned about.

Need to also double check with @igorbernstein2 if we're okay with the replication check consistency potentially taking up to 1h as well.

sandshar09: Changed single rpc timeout to 1 minute

PR #2482, opened 2025-02-11
feat: skip large rows

sarthakbhutani:
Tasks remaining -
- [ ] make changes in the read request API which skips for large rows/internally calls readLargeRowsCallable()
- [ ] expose large rows rowkeys in sidechannel/dlq/some other method - which can be exposed to client


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/LargeRowRetryAlgorithm.java
```diff
@@ -0,0 +1,102 @@
...
+   * <p>Otherwise it returns the result of {@link #shouldRetry(Throwable, Object)}.
+   */
+  @Override
+  public boolean shouldRetry(
+      @Nullable RetryingContext context, Throwable previousThrowable, ResponseT previousResponse) {
+    if (extractRetryDelay(previousThrowable) != null) {
+      // First check if server wants us to retry
+      return true;
+    }
+    if (context != null && context.getRetryableCodes() != null) {
```
mutianf, line 70: can we split this out? The comment and this block is a bit confusing. 
have a helper method called `isLargeRow` and 
```
if(isLargeRow(previousThrowable)) {
return true
}
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ServerStreamingAttemptCallable.java
```diff
@@ -221,12 +222,17 @@ public void onResponseImpl(ResponseT response) {
 
           @Override
           public void onErrorImpl(Throwable t) {
+            ((LargeReadRowsResumptionStrategy) resumptionStrategy).setLargeRowKey(t);
```
mutianf, line 225: add a check. If resumptionStrategy is largeRowResumptionStrategty then set the largeRow key.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ServerStreamingAttemptCallable.java
```diff
@@ -221,12 +222,17 @@ public void onResponseImpl(ResponseT response) {
...
             onAttemptError(t);
+            //  ToDo (@sarthakbhutani) : since this is a terminal step - we should consider dumping
```
mutianf, line 227: remove this comment. We don't want to system out the rows here.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeRowReadCallable.java
```diff
@@ -0,0 +1,87 @@
...
+import com.google.api.gax.rpc.ApiException;
+import com.google.api.gax.rpc.ResponseObserver;
+import com.google.api.gax.rpc.ServerStreamingCallable;
+import com.google.api.gax.rpc.StreamController;
+import com.google.cloud.bigtable.data.v2.stub.SafeResponseObserver;
+
+/**
+ * This callable converts the "Received rst stream" exception into a retryable {@link ApiException}.
+ */
+public final class LargeRowReadCallable<RequestT, ResponseT, RowT>
```
mutianf, line 29: Since you're modifying the ServerStreamingAttemptCallable directly we can remove this class.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeRowReadCallable.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    protected void onResponseImpl(ResponseT response) {
+      outerObserver.onResponse(response);
+    }
+
+    @Override
+    protected void onErrorImpl(Throwable t) {
+      // this has no impact
+      // if (resumptionStrategy instanceof LargeReadRowsResumptionStrategy) {
```
mutianf, line 74: I think this has no effect because we create a new strategy here: https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/RetryingServerStreamingCallable.java#L80 (which I didn't notice previously).

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+public class LargeReadRowsResumptionStrategy<RowT>
+    implements StreamResumptionStrategy<ReadRowsRequest, RowT> {
+  private final RowAdapter<RowT> rowAdapter;
+  private ByteString lastSuccessKey = ByteString.EMPTY;
+  // Number of rows processed excluding Marker row.
+  private long numProcessed;
+  private ByteString largeRowKey = ByteString.EMPTY;
+  private AtomicInteger largeRowsCount = new AtomicInteger(0);
+
+  private List<ByteString> largeRowKeys = new ArrayList<ByteString>();
```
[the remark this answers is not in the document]
sarthakbhutani: it's added here for future use.
On Request Competion - in onComplete() method we can invoke a method in resumption strategy that sends/exposes all the failed keys for the whole request together (dumpLargeRowKeys() method can be implemented)
Collection of large row keys per request should be kept somewhere, hence kept it here.

Open for suggestions.

sarthakbhutani (commented):
have resolved your comments - which are implemented in code.

unresolved comments - have not done - require a discussion - can discuss offline?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -1279,6 +1293,10 @@ public ServerStreamingCallable<Query, Row> readRowsCallable() {
     return stub.readRowsCallable();
   }
```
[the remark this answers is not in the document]
mutianf: We don't want to expose this API

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -515,6 +548,65 @@ private <ReqT, RowT> ServerStreamingCallable<ReadRowsRequest, RowT> createReadRo
...
+            readRowsSettings.getRetryableCodes());
+
+    ServerStreamingCallable<ReadRowsRequest, ReadRowsResponse> withStatsHeaders =
+        new StatsHeadersServerStreamingCallable<>(base);
+
+    // Sometimes ReadRows connections are disconnected via an RST frame. This error is transient and
+    // should be treated similar to UNAVAILABLE. However, this exception has an INTERNAL error code
+    // which by default is not retryable. Convert the exception so it can be retried in the client.
+    ServerStreamingCallable<ReadRowsRequest, ReadRowsResponse> convertException =
+        new LargeRowConvertExceptionCallable<>(withStatsHeaders);
```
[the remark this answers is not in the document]
mutianf: I don't think so? RetryAlgorithm should handle this error as retryable.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/LargeRowConvertExceptionCallable.java
```diff
@@ -0,0 +1,125 @@
+/*
```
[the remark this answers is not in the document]
mutianf: This should be handled by the retry algorithm

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+    if (lastSuccessKey.isEmpty() && largeRowKey.isEmpty()) {
+      return originalRequest;
+    }
+
+    RowSet remaining;
+
+    remaining =
+        RowSetUtil.erase(originalRequest.getRows(), lastSuccessKey, !originalRequest.getReversed());
+    if (!largeRowKeys.isEmpty()) {
+      for (ByteString largeRowKey : largeRowKeys) {
```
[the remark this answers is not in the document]
mutianf: Ok I see.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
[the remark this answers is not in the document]
sarthakbhutani: fixed

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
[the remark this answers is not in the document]
sarthakbhutani: lets discuss this offline

sarthakbhutani (commented):
added comments

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ServerStreamingAttemptCallable.java
```diff
@@ -221,12 +222,17 @@ public void onResponseImpl(ResponseT response) {
 
           @Override
           public void onErrorImpl(Throwable t) {
+            if (resumptionStrategy.getClass() == LargeReadRowsResumptionStrategy.class) {
```
mutianf, line 225:
```suggestion
            if (resumptionStrategy instanceof LargeReadRowsResumptionStrategy) {
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/gaxx/retrying/ServerStreamingAttemptCallable.java
```diff
@@ -221,12 +222,17 @@ public void onResponseImpl(ResponseT response) {
...
+              ((LargeReadRowsResumptionStrategy) resumptionStrategy).setLargeRowKey(t);
+            }
             onAttemptError(t);
           }
 
           @Override
           public void onCompleteImpl() {
             onAttemptComplete();
+            // ToDo (@sarthakbhutani) : since this is a terminal step - we should consider dumping
+            // all the large rows collection somewhere via DLQ/side channel to application layer
```
mutianf, line 235: clean up these comments

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -49,6 +49,45 @@
 public final class RowSetUtil {
   private RowSetUtil() {}
 
+  public static RowSet createSplitRanges(
+      RowSet rowSet, ByteString excludePoint, boolean fromStart) {
+    RowSet.Builder newRowSet = RowSet.newBuilder();
+
+    if (rowSet.getRowKeysList().isEmpty() && rowSet.getRowRangesList().isEmpty()) {
+      rowSet = RowSet.newBuilder().addRowRanges(RowRange.getDefaultInstance()).build();
```
mutianf, line 57: I don't think we should call `RowRange.getDefaultInstance()`? it should be     RowRange.newBuilder().setEndKeyOpen(excludePoint).build(); and RowRange.newBuilder().setStartKeyOpen(excludePoint).build();

so we're querying range (, largeRow) and (largeRow, )

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -49,6 +49,45 @@
...
+    }
+
+    // Handle point lookups
```
mutianf, line 60: I think all we need to do here is to remove the row key from the row keys list.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -1279,6 +1279,69 @@ public ServerStreamingCallable<Query, Row> readRowsCallable() {
...
+   *   // etc
+   * }
+   * }</pre>
+   *
+   * @see ServerStreamingCallable For call styles.
+   * @see Query For query options.
+   * @see com.google.cloud.bigtable.data.v2.models.Filters For the filter building DSL.
+   */
+  @InternalApi("only to be used by Bigtable beam connector")
+  public ServerStreamingCallable<Query, Row> skipLargeRowsCallable() {
```
mutianf, line 1341: I chatted with Igor, let's not expose this on the BigtableDataClient. In the beam connector we can update this line https://github.com/apache/beam/blob/4ee26065d9b4ea074d3d59d38cc4fbc5aa9dbfa5/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigtable/BigtableServiceImpl.java#L111 to create a EnahncedBigtableStub instead. I can make that change in beam.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -124,6 +134,98 @@ public void immediateRetryTest() {
     Truth.assertThat(actualResults).containsExactly("k1", "r1", "r2").inOrder();
   }
 
+  public ApiException largeRowExceptionWithTrailers(String rowKey) {
```
mutianf, line 137:
```suggestion
  public ApiException createLargeRowException(String rowKey) {
```

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -124,6 +134,98 @@ public void immediateRetryTest() {
...
+    ErrorInfo errorInfo =
+        ErrorInfo.newBuilder()
+            .setReason("LargeRowReadError")
+            .setDomain("bigtable.googleapis.com")
+            .putMetadata("rowKey", rowKey)
+            .build();
+
+    Any packedErrorInfo = Any.pack(errorInfo);
+    // ErrorDetails errorDetails = ErrorDetails.builder()
```
mutianf, line 146: nit: clean up comments

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -124,6 +134,98 @@ public void immediateRetryTest() {
     Truth.assertThat(actualResults).containsExactly("k1", "r1", "r2").inOrder();
   }
 
+  public ApiException largeRowExceptionWithTrailers(String rowKey) {
+    ErrorInfo errorInfo =
+        ErrorInfo.newBuilder()
+            .setReason("LargeRowReadError")
+            .setDomain("bigtable.googleapis.com")
+            .putMetadata("rowKey", rowKey)
```
mutianf, line 142: TODO: this should be updated once Justin's cl is merged

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -124,6 +134,98 @@ public void immediateRetryTest() {
...
+    return (ApiException)
+        (new UnavailableException(
+            new StatusRuntimeException(Status.FAILED_PRECONDITION, trailers),
+            GrpcStatusCode.of(Code.FAILED_PRECONDITION),
+            false,
+            errorDetails));
+  }
+
+  /**
+   * AdditionalTests - 1. only query for large rows - empty response? 2. 1st row is large row - then
```
mutianf, line 166: instead of the comment the test itself should describe what the test is doing. for example largeRowTestBasic could be largeRowRangesTest, and add a comment only describe what this test is testing

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -124,6 +134,98 @@ public void immediateRetryTest() {
...
+        RpcExpectation.create()
+            .expectRequestForMultipleRowRanges(rangeList)
+            .respondWith("r3", "r4"));
+
+    List<String> actualResults = getLargeRowResults(Query.create(TABLE_ID).range("r1", "r5"));
+    Truth.assertThat(actualResults).containsExactly("r1", "r3", "r4").inOrder();
+  }
+
+  @Test
+  public void largeRowTestMultipleAdhocRows() {
```
mutianf, line 193: how is this test different from the largeRowTestBasic?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -299,6 +401,15 @@ private List<String> getResults(Query query) {
     return actualValues;
   }
 
+  private List<String> getLargeRowResults(Query query) {
+    ServerStream<Row> actualRows = client.skipLargeRowsCallable().call(query);
```
mutianf, line 405: you can do 

```
List<String> actualRowKeys = client.skipLargeRowsCallable().all().call(Query.create("table"))
                    .stream().map(row -> row.getKey().toStringUtf8())
                    .collect(Collectors.toList());

return actualRowKeys; 
```

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -299,6 +401,15 @@ private List<String> getResults(Query query) {
     return actualValues;
   }
 
+  private List<String> getLargeRowResults(Query query) {
```
mutianf, line 404:
```suggestion
  private List<String> getSkipLargeRowsResults(Query query) {
```

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -124,6 +134,98 @@ public void immediateRetryTest() {
...
+    //     .build();
+
+    ErrorDetails errorDetails =
+        ErrorDetails.builder().setRawErrorMessages(ImmutableList.of(packedErrorInfo)).build();
+
+    Metadata trailers = new Metadata();
+    byte[] status =
+        com.google.rpc.Status.newBuilder().addDetails(Any.pack(errorInfo)).build().toByteArray();
+    trailers.put(ERROR_DETAILS_KEY, status);
+    return (ApiException)
```
mutianf, line 157: You don't need to cast, UnavailableException is a ApiException https://github.com/googleapis/sdk-platform-java/blob/main/gax-java/gax/src/main/java/com/google/api/gax/rpc/UnavailableException.java#L36

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -322,6 +433,8 @@ public void readRows(
       }
       if (expectedRpc.statusCode.toStatus().isOk()) {
         responseObserver.onCompleted();
+      } else if (expectedRpc.statusException != null) {
```
mutianf, line 436: I don't think you need to create a different list. Just add them to expectedRpc.exception. Because your largeRowException is returning ApiExceptions

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -355,6 +469,58 @@ RpcExpectation expectRequest(String... keys) {
...
+      return this;
+    }
+
+    /**
+     * Build Row Range
+     *
+     * @param range
+     * @return
+     */
+    RowRange rangeBuilder(Range<String> range) {
```
mutianf, line 486: I'm a bit confused, whats this method for? you can use `Query.create(..).range(Range.ByteStringRange.create());` instead, which includes all the start / end open / close combinations https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/Range.java#L69-L103

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -124,6 +134,98 @@ public void immediateRetryTest() {
...
+            new StatusRuntimeException(Status.FAILED_PRECONDITION, trailers),
+            GrpcStatusCode.of(Code.FAILED_PRECONDITION),
+            false,
+            errorDetails));
+  }
+
+  /**
+   * AdditionalTests - 1. only query for large rows - empty response? 2. 1st row is large row - then
+   * get other responses 3. last row is large row 4. multiple adhoc large rows 5. continous large
+   * rows
```
mutianf, line 168: I also don't think everything described in this comment is added. I think they're all valuable and we should add them.

Additionally, we should add a few more tests
1. Test row keys logic is correct. Query.create().rowKey(key1).rowKey(k2).rowKey(largeRow). 
2. Test the row limit behavior is still correct. 
3. Test reverse scan. `Query.create(..).reversed(true);` is set. `reversed(true)` will return rows in the reversed order, so instead of returning rows like "r1, r2, r3", it'll be "r3, r2, r1".

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,67 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+      // either set = or set the range's start & end key acc to you
+    }
+
+    if (fromStart) {
+      // range start is on or left of the split
+      if (StartPoint.extract(range).compareTo(new StartPoint(split, true)) < 0) {
+        RowRange beforeSplitKeyRange =
+            range
+                .toBuilder()
+                .setStartKeyOpen(
```
mutianf, line 193: we don't want to touch the start key. just do `range.setEndKeyOpen(split)`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,67 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+                    range.getStartKeyOpen().isEmpty()
+                        ? range.getStartKeyClosed()
+                        : range.getStartKeyOpen())
+                .setEndKeyOpen(split)
+                .build();
+        rowRangesList.add(beforeSplitKeyRange);
+        // RowRange afterSplitKeyRange = range.toBuilder().setStartKeyOpen(split).build();
+      }
+      if (EndPoint.extract(range).compareTo(new EndPoint(split, true)) > 0) {
```
mutianf, line 202: this statement is already covered by the previous block (line 174 to 176). so it's redundant here. we should always add RowRange afterSplitKeyRange = range.toBuilder().setStartKeyOpen(split);

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,67 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+            range
+                .toBuilder()
+                .setStartKeyOpen(
+                    range.getStartKeyOpen().isEmpty()
+                        ? range.getStartKeyClosed()
+                        : range.getStartKeyOpen())
+                .setEndKeyOpen(split)
+                .build();
+        rowRangesList.add(beforeSplitKeyRange);
+        // RowRange afterSplitKeyRange = range.toBuilder().setStartKeyOpen(split).build();
```
mutianf, line 200: I think this is correct and we should uncomment it and remove the block from 202 - 205

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,67 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+      }
+    } else {
+      // range end is on or right of the split
+      if (EndPoint.extract(range).compareTo(new EndPoint(split, true)) > 0) {
+        // end key open would be already present as large read success key
+        RowRange afterSplitKeyRange =
+            range
+                .toBuilder()
+                .setStartKeyOpen(split)
+                .setEndKeyOpen(
```
mutianf, line 214: I don't think we should touch the end key.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,67 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+                    range.getEndKeyOpen().isEmpty()
+                        ? range.getEndKeyClosed()
+                        : range.getEndKeyOpen())
+                .build();
+        rowRangesList.add(afterSplitKeyRange);
+      }
+      if (StartPoint.extract(range).compareTo(new StartPoint(split, true)) < 0) {
```
mutianf, line 221: same as above

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,67 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
     return newRange.build();
   }
 
+  public static List<RowRange> splitRange(RowRange range, ByteString split, boolean fromStart) {
```
mutianf, line 169: Maybe a better name would be eraseKey or something

mutianf, line 169: and add a comment saying this method will erase splitPoint from the rowRange

mutianf:
We decided to use base64 so let's add the decoding part in the resumption strategy and unit tests. Integration test can wait on server changes.

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/LargeRowIT.java
```diff
@@ -73,4 +88,257 @@ public void testWriteRead() throws Exception {
...
+
+    AccumulatingObserver observer2 = new AccumulatingObserver();
+    Query query2 = Query.create(tableId).range("r1", "r5");
+    client.skipLargeRowsCallable().call(query2, observer2);
+    observer2.awaitCompletion();
+    assertThat(observer2.responses).containsExactly(expectedRow1, expectedRow4);
+  }
+
+  @Test
+  public void fastLargeRowTestWhenRowsAreAlreadyCreated() throws Throwable {
```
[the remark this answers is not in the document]
sarthakbhutani: removed

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/it/LargeRowIT.java
```diff
@@ -73,4 +88,257 @@ public void testWriteRead() throws Exception {
...
+    // --- large row creation START----
+    byte[] largeValueBytes = new byte[9 * 1024 * 1024];
+    ByteString largeValue = ByteString.copyFrom(largeValueBytes);
+
+    int threadCount = 5;
+    int offset = 0;
+    int iterations = 80;
+    ExecutorService executor = Executors.newFixedThreadPool(4);
+
+    for (int i = offset; i < threadCount + offset; i++) {
```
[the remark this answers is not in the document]
sarthakbhutani: tried this way.
in all the test cases where i assumed r2 to be the large row, it was being returned (idk why?)

hence, keeping this async code again

sarthakbhutani (commented):
have reverted back to comments.
let's discuss the open points offline?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,108 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+    }
+
+    // split out of range or on edge of the range
+    if (fromStart) {
+      // range end is on or left of the split
+      if (EndPoint.extract(range).compareTo(new EndPoint(split, true)) < 0) {
+        rowRangesList.add(range);
+        return rowRangesList;
+      }
+      else if(EndPoint.extract(range).compareTo(new EndPoint(split, true)) == 0){
```
mutianf, line 200: nit: add a comment:
// range end is on the split

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,108 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+        range = range.toBuilder().setEndKeyOpen(split).build();
+        rowRangesList.add(range);
+        return rowRangesList;
+      }
+      //range start is on the left or on the split
+      else if(StartPoint.extract(range).compareTo(new StartPoint(split,true))>0){
+        rowRangesList.add(range);
+        return rowRangesList;
+      }
+      else if(StartPoint.extract(range).compareTo(new StartPoint(split,true))==0){
```
mutianf, line 210: nit: add a comment:
// range start is on the split

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,108 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+        range = range.toBuilder().setStartKeyOpen(split).build();
+        rowRangesList.add(range);
+        return rowRangesList;
+      }
+
+    } else {
```
mutianf, line 216: I don't think `fromStart` matters here, because you're just removing a point in the row ranges. And if you compare the logic between these 2 blocks they're identical. So you can remove this check. And add a comment saying why it doesn't matter.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,108 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+      }
+      else if(EndPoint.extract(range).compareTo(new EndPoint(split,true))==0){
+        range = range.toBuilder().setEndKeyOpen(split).build();
+        rowRangesList.add(range);
+        return rowRangesList;
+      }
+    }
+
+    // split is b/w the range
+    if (fromStart) {
```
mutianf, line 240: Same comment about `fromtStart` apply here too.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,108 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+      if (StartPoint.extract(range).compareTo(new StartPoint(split, true)) < 0) {
+        RowRange beforeSplitKeyRange =
+            range
+                .toBuilder()
+                .setEndKeyOpen(split)
+                .build();
+        rowRangesList.add(beforeSplitKeyRange);
```
mutianf, line 247: how about the range (split, originalEnd] ? I think you need to add that too.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,108 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+      }
+      if (EndPoint.extract(range).compareTo(new EndPoint(split, true)) > 0) {
+        RowRange afterSplitKeyRange = range.toBuilder().setStartKeyOpen(split).build();
+        rowRangesList.add(afterSplitKeyRange);
```
mutianf, line 251: you need to add [originalStart, split)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -49,6 +52,44 @@
...
+
+    if (rowSet.getRowKeysList().isEmpty() && rowSet.getRowRangesList().isEmpty()) {
+      // querying range (, excludePoint) and (excludePoint, )
+      newRowSet.addRowRanges(RowRange.newBuilder().setEndKeyOpen(excludePoint).build());
+      newRowSet.addRowRanges(RowRange.newBuilder().setStartKeyOpen(excludePoint).build());
+    }
+
+    // Handle point lookups
+    if(!rowSet.getRowKeysList().isEmpty()){
+      List<ByteString> mutableList = new ArrayList<ByteString>();
```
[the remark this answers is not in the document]
sarthakbhutani: that doesnt work - it's a mutable list.
throws UnsupportedException

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,108 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
+   * This method erases the {@code #split} range from the range
+   * @param range
+   * @param split
+   * @param fromStart
+   * @return
+   */
+  public static List<RowRange> eraseKeyFromRange(RowRange range, ByteString split, boolean fromStart) {
+    List<RowRange> rowRangesList = new ArrayList<RowRange>();
+
+    //edge case of [split,split],[split,split),(split,split]
```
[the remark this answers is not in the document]
sarthakbhutani: discussed over call

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
sarthakbhutani: will check & update

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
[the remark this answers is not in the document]
mutianf: Do something like .stream()
                .filter(s -> !s.equals(".."))
                .collect(Collectors.toList()); 
instead

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -125,6 +166,108 @@ private static RowRange truncateRange(RowRange range, ByteString split, boolean
...
   }
 
+  /**
+   * This method erases the {@code #split} range from the range
+   * @param range
+   * @param split
+   * @param fromStart
+   * @return
+   */
+  public static List<RowRange> eraseKeyFromRange(RowRange range, ByteString split, boolean fromStart) {
```
mutianf, line 176: You can simplify this logic: 

```
    public static List<RowRange> splitOnLargeRowKey(RowRange range, ByteString largeRowKey) {
        List<RowRange> rowRanges = new ArrayList<>();

        ByteString startKey = StartPoint.extract(range).value;
        ByteString endKey = EndPoint.extract(range).value;

        // if end key is on the left of large row key, don't split
        if (ByteStringComparator.INSTANCE.compare(endKey, largeRowKey) < 0) {
            rowRanges.add(range);
            return rowRanges;
        }

        // if start key is on the right of the large row key, don't split
        if (ByteStringComparator.INSTANCE.compare(startKey, largeRowKey) > 0) {
            rowRanges.add(range);
            return rowRanges;
        }

        // if start key is on the left of the large row key, set the end key to be large row key open
        if (ByteStringComparator.INSTANCE.compare(startKey, largeRowKey) < 0) {
            RowRange beforeSplit = range
                    .toBuilder()
                    .setEndKeyOpen(largeRowKey)
                    .build();
            rowRanges.add(beforeSplit);
        }

        // if the end key is on the right of the large row key, set the start key to be large row key open
        if (ByteStringComparator.INSTANCE.compare(endKey, largeRowKey) > 0) {
            RowRange afterSplit = range
                    .toBuilder()
                    .setStartKeyOpen(largeRowKey)
                    .build();
            rowRanges.add(afterSplit);
        }
       return rowRanges;
    }
```

This should cover all edge cases.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -49,6 +52,44 @@
 public final class RowSetUtil {
   private RowSetUtil() {}
 
+  public static RowSet createSplitRanges(
```
mutianf, line 55: I think you should also resume from the last seen row key.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+  // Number of rows processed excluding Marker row.
+  private long numProcessed;
+  private ByteString largeRowKey = ByteString.EMPTY;
+  private final AtomicInteger largeRowsCount = new AtomicInteger(0);
+
+  private ReadRowsRequest originalRequest;
+
+  private ReadRowsRequest lastModifiedRequestOnTimeOfError;
+
+  private RowSet previousFailedRequestRowset = null;
```
mutianf, line 61: See my comment in RowSetUtil. We can remove `previousFailedRequestRowset`. Since we're updating lastSuccessKey, we'll always update the request to remove everything before lastSuccessKey first. So all the largeRowKeys we've seen before should be excluded from the next request. 

For example:

Original request, row range [r1, r10]. Large row keys: r3, r7.

First attempt: read everything up to r2 successfully, failed on r3. 
Resumption strategy will return new range (r2, r3), (r3, r10].

Second attempt: read everything up to r6 successfully, failed on r7.
Resumption strategy will use the original request [r1, r10],  and update the range to be (r6, r7), (r7, r10].

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -26,6 +26,9 @@
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ComparisonChain;
 import com.google.protobuf.ByteString;
+// import com.google.protobuf.Message;
```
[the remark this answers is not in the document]
mutianf: Still not cleaned up :(

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -515,6 +521,96 @@ private <ReqT, RowT> ServerStreamingCallable<ReadRowsRequest, RowT> createReadRo
...
+   * internally. The chain will:
+   *
+   * <ul>
+   *   <li>Convert a {@link Query} into a {@link com.google.bigtable.v2.ReadRowsRequest}.
+   *   <li>Dispatch the RPC with {@link ReadRowsRequest}.
+   *   <li>Upon receiving the response stream, it will merge the {@link
+   *       com.google.bigtable.v2.ReadRowsResponse.CellChunk}s in logical rows. The actual row
+   *       implementation can be configured in by the {@code rowAdapter} parameter.
+   *   <li>Add bigtable tracer for tracking bigtable specific metrics.
+   *   <li>Retry/resume on failure (retries for retryable error codes, connection errors & large row
```
mutianf, line 535:
```suggestion
   *   <li>Retry/resume on failure (retries for retryable error codes, connection errors and skip large row
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -515,6 +521,96 @@ private <ReqT, RowT> ServerStreamingCallable<ReadRowsRequest, RowT> createReadRo
...
+   *       errors)
```
mutianf, line 536:
```suggestion
   *       keys)
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStub.java
```diff
@@ -515,6 +521,96 @@ private <ReqT, RowT> ServerStreamingCallable<ReadRowsRequest, RowT> createReadRo
...
+            readRowsSettings.getRetryableCodes());
+
+    ServerStreamingCallable<ReadRowsRequest, ReadRowsResponse> withStatsHeaders =
+        new StatsHeadersServerStreamingCallable<>(base);
+
+    // Sometimes ReadRows connections are disconnected via an RST frame. This error is transient and
+    // should be treated similar to UNAVAILABLE. However, this exception has an INTERNAL error code
+    // which by default is not retryable. Convert the exception so it can be retried in the client.
+    ServerStreamingCallable<ReadRowsRequest, ReadRowsResponse> convertException =
+        new LargeRowConvertExceptionCallable<>(withStatsHeaders);
```
mutianf, line 565: This is still not cleaned up..

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -1206,6 +1206,7 @@ public EnhancedBigtableStubSettings build() {
               // GrpcHeaderInterceptor treats the `user-agent` as a magic string
               .put("user-agent", "bigtable-java/" + Version.VERSION)
               .put("bigtable-features", new String(encodedFlags, StandardCharsets.UTF_8))
+              .put("cookie", "TR=T=EMziZs0CRy4:X=nsxV8:S=8VDZW1q3Kanhhy8n")
```
mutianf, line 1209: remove things used for debugging :)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/LargeRowConvertExceptionCallable.java
```diff
@@ -0,0 +1,125 @@
...
+ * Copyright 2021 Google LLC
```
mutianf, line 2: Please remove this file

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+      numProcessed = numProcessed + 1;
+    }
+    return throwable;
+  }
+
+  private String extractLargeRowKey(Throwable t) {
+    if (t instanceof ApiException
+        && ((ApiException) t).getReason() != null
+        && ((ApiException) t).getReason().equals("LargeRowReadError")) {
+      String rowKey = ((ApiException) t).getMetadata().get("rowKey");
```
mutianf, line 106: this is updated in the latest cl. the key is "rowKeyBase64Encoded". And we should decode it.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
```diff
@@ -0,0 +1,169 @@
...
+    if (!rowAdapter.isScanMarkerRow(response)) {
+      // Only real rows count towards the rows limit.
+      numProcessed++;
+    }
+    return response;
+  }
+
+  public Throwable processError(Throwable throwable) {
+    String rowKeyExtracted = extractLargeRowKey(throwable);
+    if (rowKeyExtracted != null) {
```
mutianf, line 94: Let's log the large rows here. You can add a logger to this class with 

```
private static final Logger LOGGER = Logger.getLogger(LargeReadRowsResumptionStratigy.class.getName());
```

and here add 

```
LOGGER.wraning("skipping large row " + rowKeyExtracted);
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
sarthakbhutani: we needed to keep the prev request, because, for the use case of multiple large row keys together.
if we dont keep the prev request, if the prev request failed because of large-row and the next request also failed because of large-row, in this request, we will remove the 2nd large-row key & the prev one would be there, this would keep failing

ex - 
request [r1,r4]
r1 - success key
r2 - large-row key
r3 - large-row key
r4 - large-row key


original request -> [r1,r4]
r1 read, r2 fails -> req becomes (r1,r2),(r2,r4]
r3 fails, req becomes -> (r1,r3), (r3,r4] -> which will now fail for r2 again (if the prev request or prev failed row keys are not cached)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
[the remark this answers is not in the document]
sarthakbhutani: i didnt push the code after your last review, since it got pretty late for me the other night

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/RowSetUtil.java
```diff
@@ -49,6 +50,38 @@
...
+      newRowSet.addRowRanges(RowRange.newBuilder().setEndKeyOpen(excludePoint).build());
+      newRowSet.addRowRanges(RowRange.newBuilder().setStartKeyOpen(excludePoint).build());
+    }
+
+    // remove large row key from point reads
+    rowSet.getRowKeysList().stream().filter(k -> !k.equals(excludePoint)).forEach(newRowSet::addRowKeys);
+
+    // Handle ranges
+    for (RowRange rowRange : rowSet.getRowRangesList()) {
+      // List<RowRange> afterSplit = eraseKeyFromRange(rowRange, excludePoint, fromStart);
```
mutianf, line 72: nit: clean up this comment

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
mutianf:
```suggestion
      String rowKey = ((ApiException) t).getMetadata().get("rowKeyBase64Encoded");
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
```diff
@@ -0,0 +1,173 @@
...
+    return throwable;
+  }
+
+  private String extractLargeRowKey(Throwable t) {
+    if (t instanceof ApiException
+        && ((ApiException) t).getReason() != null
+        && ((ApiException) t).getReason().equals("LargeRowReadError")) {
+      String rowKey = ((ApiException) t).getMetadata().get("rowKey");
+      byte[] decodedBytes = Base64.getDecoder().decode(rowKey);
+      return new String(decodedBytes);
```
mutianf, line 112: We should return ByteString directly 
`ByteString.copyFrom(decodedBytes)`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
```diff
@@ -0,0 +1,173 @@
...
+public class LargeReadRowsResumptionStrategy<RowT>
+    extends BigtableStreamResumptionStrategy<ReadRowsRequest, RowT> {
+  private final RowAdapter<RowT> rowAdapter;
+  private ByteString lastSuccessKey = ByteString.EMPTY;
+  // Number of rows processed excluding Marker row.
+  private long numProcessed;
+  private ByteString largeRowKey = ByteString.EMPTY;
+  private final AtomicInteger largeRowsCount = new AtomicInteger(0);
+
+  private ReadRowsRequest originalRequest;
```
mutianf, line 58: pretty sure this is unused. Can we remove it?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
```diff
@@ -0,0 +1,173 @@
...
+    if (remaining == null
+        || (originalRequest.getRowsLimit() > 0 && originalRequest.getRowsLimit() == numProcessed)) {
+      return ReadRowsRetryCompletedCallable.FULFILLED_REQUEST_MARKER;
+    }
+
+    Builder builder = originalRequest.toBuilder().setRows(remaining);
+
+    if (originalRequest.getRowsLimit() > 0) {
+      Preconditions.checkState(
+          originalRequest.getRowsLimit() > numProcessed + largeRowsCount.get(),
```
mutianf, line 166: I think this check here we don't need to include the large rows count because we are already incrementing numProcessed in processError

sarthakbhutani (commented):
add comments

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/readrows/LargeReadRowsResumptionStrategy.java
sarthakbhutani: remove largerowcounts counter

mutianf: ADd we can probably remove the largeRowCount

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -124,6 +136,215 @@ public void immediateRetryTest() {
...
+        false,
+        errorDetails));
+  }
+
+  /**
+   * This tests if in a read rows request RowRange includes large rows, those rows are omitted in
+   * the response.
+   */
+  @Test
+  public void readRowRangeWithSkippingLargeRows() {
```
mutianf, line 168: break these out into 2 seprate test cases

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -160,6 +161,81 @@ public ApiException createLargeRowException(String rowKey) {
...
+            .expectRequest(Range.closedOpen("r4", "r7"))
+            .expectRowLimit(2)
+            .respondWith("r4", "r5"));
+
+    actualResults = getSkipLargeRowsResults(Query.create(TABLE_ID).range("r4", "r7").limit(2));
+    Truth.assertThat(actualResults).containsExactly("r4", "r5").inOrder();
+  }
+
+  @Test
+  public void readRowsTestWithMultipleLargeRowsTest() {
```
mutianf, line 194: This test case look identical to readRowsWithLimitSkippingLargeRowsTest?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
```diff
@@ -160,6 +161,81 @@ public ApiException createLargeRowException(String rowKey) {
...
+            .respondWith("r1", "r4")
+            .respondWithException(Code.INTERNAL, largeRowExceptionWithTrailersR7));
+    service.expectations.add(RpcExpectation.create().expectRequest("r8").respondWith("r8"));
+
+    actualResults =
+        getSkipLargeRowsResults(
+            Query.create(TABLE_ID).rowKey("r1").rowKey("r7").rowKey("r4").rowKey("r8"));
+    Truth.assertThat(actualResults).containsExactly("r1", "r4", "r8").inOrder();
+  }
+
```
mutianf, line 238: Can we also add a test case for reverse scan?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/readrows/ReadRowsRetryTest.java
sarthakbhutani: it also has limit

mutianf (approved):

PR #2508, opened 2025-03-10
feat(bigtable): Add support for data APIs for materialized views

ron-gal:
- [x] Rollback plan is reviewed and LGTMed
- [x] All new data plane features have a completed end to end testing plan

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/NameUtil.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/NameUtil.java
@@ -106,6 +106,41 @@ public class NameUtil {
     return matcher.group(3);
   }
 
+  /** A helper to convert fully qualified tableName andauthorizedViewName to a {@link TargetId} */
+  public static TargetId extractTargetId(
+      @Nonnull String tableName, @Nonnull String authorizedViewName) {
+    if (tableName.isEmpty() && authorizedViewName.isEmpty()) {
+      throw new IllegalArgumentException(
+          "Either table name, authorized view name or materialized view name must be specified. Table name: "
+              + tableName
+              + ", authorized view name: "
+              + authorizedViewName);
+    }
+    int names = 0;
+    if (!tableName.isEmpty()) {
+      ++names;
+    }
+    if (!authorizedViewName.isEmpty()) {
+      ++names;
+    }
+    if (names > 1) {
+      throw new IllegalArgumentException(
+          "Only one of table name and authorized view name can be specified at the same time. Table name: "
+              + tableName
+              + ", authorized view name: "
+              + authorizedViewName);
+    }
+
+    if (!tableName.isEmpty()) {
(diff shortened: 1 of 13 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/models/BulkMutation.java
```diff
@@ -155,8 +155,9 @@ public static BulkMutation fromProto(@Nonnull MutateRowsRequest request) {
     String tableName = request.getTableName();
     String authorizedViewName = request.getAuthorizedViewName();
 
+    // Materialized Views are read only entities.
     BulkMutation bulkMutation =
-        BulkMutation.create(NameUtil.extractTargetId(tableName, authorizedViewName));
+        BulkMutation.create(NameUtil.extractTargetId(tableName, authorizedViewName, ""));
```
andre-sampaio, line 160: Since we have a very clear scenario for which it never makes sense to pass the third parameter (any kind of mutation), maybe create an overload for `extractTargetId` with only the first 2 parameters

ron-gal: Good idea, done!

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/NameUtil.java
```diff
@@ -88,19 +96,36 @@ public static String extractAuthorizedViewIdFromAuthorizedViewName(
...
+    Matcher matcher = MATERIALIZED_VIEW_PATTERN.matcher(fullMaterializedViewName);
+    if (!matcher.matches()) {
+      throw new IllegalArgumentException(
+          "Invalid materialized view name: " + fullMaterializedViewName);
+    }
+    return matcher.group(3);
+  }
+
+  /** A helper to convert fully qualified tableName andauthorizedViewName to a {@link TargetId} */
   public static TargetId extractTargetId(
```
andre-sampaio, line 110: Now this overload can just proxy the request to the other one so we don't have the logic duplicated

andre-sampaio:
Regarding the `clirr` error, you can add the interface here: google-cloud-bigtable/clirr-ignored-differences.xml

That's a check for possible breaking changes, but we consider `TargetId` an internal entity, so it is ok to break compatibility here

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/NameUtil.java
ron-gal: Done

ron-gal:
> Regarding the `clirr` error, you can add the interface here: google-cloud-bigtable/clirr-ignored-differences.xml
> 
> That's a check for possible breaking changes, but we consider `TargetId` an internal entity, so it is ok to break compatibility here

Done

andre-sampaio (approved):

PR #2509, opened 2025-03-11
feat(large-row-skip): added large-row-skip-callable with configurable rowadapter

sarthakbhutani:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
@@ -1361,6 +1361,43 @@ public class BigtableDataClient implements AutoCloseable {
     return stub.createReadRowsCallable(rowAdapter);
   }
 
+
+  /**
+   * Streams back the results of the query. This callable allows for customization of the logical
+   * representation of a row. It's meant for advanced use cases.
+   *
+   * <p>Sample code:
+   *
+   * <pre>{@code
+   * try (BigtableDataClient bigtableDataClient = BigtableDataClient.create("[PROJECT]", "[INSTANCE]")) {
+   *   String tableId = "[TABLE]";
+   *
+   *   Query query = Query.create(tableId)
+   *          .range("[START KEY]", "[END KEY]")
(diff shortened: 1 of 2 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -1361,6 +1361,43 @@ public <RowT> ServerStreamingCallable<Query, RowT> readRowsCallable(RowAdapter<R
     return stub.createReadRowsCallable(rowAdapter);
   }
 
+  /**
+   * Streams back the results of the query. This callable allows for customization of the logical
```
justinuang, line 1365: Mention that this skips large rows

justinuang (approved):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/BigtableDataClient.java
```diff
@@ -1361,6 +1361,43 @@ public <RowT> ServerStreamingCallable<Query, RowT> readRowsCallable(RowAdapter<R
     return stub.createReadRowsCallable(rowAdapter);
   }
 
+  /**
+   * Streams back the results of the query skipping the large-rows. This callable allows for customization of the logical
+   * representation of a row. It's meant for advanced use cases.
```
andre-sampaio, line 1366: I know we already have the `InternalApi` decorator, but can we also make it explicit on the method docs that this is an internal API, it is subject to breaking changes and should not be relied on by user code?

Just want to make sure this is very explicit for now

andre-sampaio (approved):

andre-sampaio:
Ah also please make sure the conventionalcommits check passes (I suspect it may just be missing a space after `feat:`)

sarthakbhutani:
feat: is present in the first commit
I've formatted the code & added the comment as suggested

andre-sampaio:
> feat: is present in the first commit I've formatted the code & added the comment as suggested

You should be able to just change the title of the PR, that's the message that we will use when we squash + merge

sarthakbhutani:
it already contains feat: what needs to be changed?

andre-sampaio:
or even better:

`feat(large-row-skip): added large-row-skip-callable with configurable rowadapter`

PR #2515, opened 2025-03-12
chore: use 2 different exporter instances for internal and external metrics

igorbernstein2:
This is necessary to unblock upcoming changes that will track additional metrics and a new unified BigtableClient monitored resource.

The primary change was to extract metric data creation from BigtableCloudMonitoringExporter into TimeSeriesConverter. This removes the code duplication in exportBigtableResourceMetrics & exportApplicationResourceMetrics. But more importantly it allows us to have to 2 separate instances of BigtableCloudMonitoringExporter - one for internal metrics and another for public metrics. This will be needed in follow up PRs that will change the monitored resource that internal metrics use - the new monitored resource is incompatible CustomOpenTelemetryMetricsProvider that is needed for public metrics

Change-Id: I3a051f1faccaba1fc17e4151d1ed85ff8e5952f6

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
@@ -40,9 +40,10 @@ import com.google.auth.Credentials;
 import com.google.cloud.monitoring.v3.MetricServiceClient;
 import com.google.cloud.monitoring.v3.MetricServiceSettings;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Supplier;
-import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.util.concurrent.MoreExecutors;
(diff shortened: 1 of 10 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -123,7 +134,7 @@ static Map<String, List<TimeSeries>> convertToBigtableTimeSeries(
       }
 
       for (PointData pd : metricData.getData().getPoints()) {
-        String projectId = getProjectId(pd);
+        ProjectName projectId = getProjectName(pd);
```
andre-sampaio, line 137: NIT: `:%s/projectId/projectName`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -107,14 +118,14 @@ static String getDefaultTaskValue() {
     return "java-" + UUID.randomUUID() + jvmName;
   }
 
-  static String getProjectId(PointData pointData) {
-    return pointData.getAttributes().get(BIGTABLE_PROJECT_ID_KEY);
+  static ProjectName getProjectName(PointData pointData) {
+    return ProjectName.of(pointData.getAttributes().get(BIGTABLE_PROJECT_ID_KEY));
   }
 
   // Returns a list of timeseries by project id
```
andre-sampaio, line 125: NIT: `:%s/project id/project name`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableCloudMonitoringExporter.java
```diff
@@ -94,43 +96,26 @@ public final class BigtableCloudMonitoringExporter implements MetricExporter {
   // https://cloud.google.com/monitoring/quotas#custom_metrics_quotas.
   private static final int EXPORT_BATCH_SIZE_LIMIT = 200;
 
-  private final MetricServiceClient client;
+  private static String exporterName;
```
andre-sampaio, line 99: Maybe my java is worse than I thought, but won't this being static break when we instantiate 2 exporters?

```
MetricExporter publicExporter =
        BigtableCloudMonitoringExporter.create(
            "bigtable metrics", ...);
    MetricExporter internalExporter =
        BigtableCloudMonitoringExporter.create(
            "application metrics", ...);
// Now exporterName is "application metrics" for both exporters
```

andre-sampaio: Now that I think about it: won't it be better if we change all  / almost all of the existing static members to non-static?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
igorbernstein2: fixed

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
igorbernstein2: fixed

PR #2562, opened 2025-04-23
fix: fix retry info algorithm setting

mutianf:
Fix a bug in RetryInfo retry algorithm. 

In the following scenario: error retry delay = 1, error retry delay = 2, error with no retry delay. Currently the client did not set retry delay duration which is default to 0 for the first attempt. This will cause all the subsequent attempts from the same error to not have any retry delay if the retry delay is not present.

The fix sets retry delay duration to be the retry delay from the last attempt. We'll decide if we want to change this behavior later.

Conformance test currently assumes that the client will reset the retry delay. So also disabling the conformance test for now.


tpc-test/src/main/java/Main.java
andre-sampaio, line 1: I think these are unrelated?

Same for the other files (except `RetryInfoRetryAlgorithm` and `known_failures`)

andre-sampaio:
Is it feasible to add a test for this scenario? e.g. the actual behaviour of the retry when getting `error retry delay = 1` -> `error retry delay = 2` -> `error with no retry delay`

mutianf:
> Is it feasible to add a test for this scenario? e.g. the actual behaviour of the retry when getting `error retry delay = 1` -> `error retry delay = 2` -> `error with no retry delay`

Yeah, in the conformance test once we decide what to do!

andre-sampaio (approved):

andre-sampaio:
lgtm!

PR #2577, opened 2025-04-29
fix: Set a default mtls endpoint value

lqiu96:
Set a default mTLSEndpoint value to pass to Gax.

For this draft PR, I also pulled in the shared-deps PR changes that introduced these DirectPath issues.

### DIFF
--- a/.github/workflows/unmanaged_dependency_check.yaml
+++ b/.github/workflows/unmanaged_dependency_check.yaml
@@ -14,6 +14,6 @@ jobs:
       shell: bash
       run: .kokoro/build.sh
     - name: Unmanaged dependency check
-      uses: googleapis/sdk-platform-java/java-shared-dependencies/unmanaged-dependency-check@google-cloud-shared-dependencies/v3.44.0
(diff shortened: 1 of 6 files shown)


lqiu96:
Hi @mutianf and @andre-sampaio, this is the proposed change that I was thinking about doing inside the BigTable repo. However, I just had some hesitations regarding existing mTLS behavior in BigTable.

BigTable's EnhancedStubSettings wrappers don't directly extend from StubSettings: https://github.com/googleapis/java-bigtable/blob/7a46352124b8de9975abb51831a7ab4c7c88f598/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java#L104

It looks to be missing the mtlsEndpoint that we set for all GAPICs in the generated StubSettings class: https://github.com/googleapis/java-bigtable/blob/7a46352124b8de9975abb51831a7ab4c7c88f598/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/BigtableStubSettings.java#L462. This is the assumption that we made regarding all Java SDK libraries (there would be a mTLS endpoint value set to the StubSettings).

Let us know what you think about this and/ or if this has an unintended side effects. Thanks!

mutianf (approved):

PR #2619, opened 2025-07-02
feat(bigtable): Add schema bundle support

ron-gal:


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CreateSchemaBundleRequest.java
```diff
@@ -0,0 +1,107 @@
...
+    this.tableId = tableId;
+    requestBuilder.setSchemaBundleId(schemaBundleId);
+  }
+
+  /**
+   * Sets the proto schema for this schema bundle.
+   *
+   * @see SchemaBundleType for details.
+   */
+  public CreateSchemaBundleRequest setProtoSchema(@Nonnull String protoSchemaFile)
```
andre-sampaio, line 67: Is it ever valid for a request not to have a protoSchema set? Otherwise we may want to either make it a required argument for creating a `CreateSchemaBundleRequest` or validate it before converting it to proto

It also feels like requiring a file path is overly restrictive (what if the file is on gcs / some database / etc). Can we accept a string and provide a utility method to read from a file instead if we think that's important (e.g. `setProtoSchemaFromFile(Path file)`)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/SchemaBundle.java
```diff
@@ -0,0 +1,99 @@
...
+    Preconditions.checkArgument(!proto.getName().isEmpty(), "SchemaBundle must have a name");
+    Preconditions.checkArgument(
+        proto.hasProtoSchema(), "Schemabundle must have a proto_schema field");
+    this.proto = proto;
+  }
+
+  /** Gets the schema bundle's id. */
+  public String getId() {
+    // Constructor ensures that name is not null.
+    SchemaBundleName fullName = SchemaBundleName.parse(proto.getName());
```
andre-sampaio, line 51: NIT: we can store the parsed SchemaBundleName as an instance member in the constructor

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/UpdateSchemaBundleRequest.java
```diff
@@ -0,0 +1,140 @@
...
+    Preconditions.checkNotNull(schemaBundleId, "schemaBundleId must be set");
+    Preconditions.checkNotNull(requestBuilder, "proto builder must be set");
+
+    this.tableId = tableId;
+    this.schemaBundleId = schemaBundleId;
+    this.requestBuilder = requestBuilder;
+  }
+
+  /** Changes the deletion protection of an existing schema bundle. */
+  public UpdateSchemaBundleRequest setProtoSchema(String protoSchemaFile) throws IOException {
```
andre-sampaio, line 82: Same comment as from CreateSchemaBundleRequest

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/SchemaBundle.java
ron-gal: Done

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/CreateSchemaBundleRequest.java
ron-gal: We do the same for views and their queries, so I don't think we want to move the protoSchemaFile to the ctor, for consistency.

We could support accept the file contents as byte array, and have the caller get these bytes from anywhere.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/UpdateSchemaBundleRequest.java
ron-gal: Same answer :)

andre-sampaio (approved):

PR #2685, opened 2025-10-06
Add CSM for batch write flow control

kongweihan:


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/RateLimitingServerStreamingCallable.java
```diff
@@ -155,22 +158,26 @@ public double getRate() {
...
       if (now.isBefore(nextTime)) {
+        bigtableTracer.addBatchWriteFlowControlFactor(cappedFactor, status, false);
         return;
       }
 
       Instant newNextTime = now.plusSeconds(period.getSeconds());
 
       if (!nextRateUpdateTime.compareAndSet(nextTime, newNextTime)) {
         // Someone else updated it already.
+        bigtableTracer.addBatchWriteFlowControlFactor(cappedFactor, status, false);
```
mutianf, line 180: Do we want to update the metric? the qps is not updated here.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/RateLimitingServerStreamingCallable.java
```diff
@@ -260,11 +269,11 @@ protected void onCompleteImpl() {
       outerObserver.onComplete();
     }
 
-    private void updateQps(double factor, Duration period) {
+    private void updateQps(double factor, Duration period, @Nullable Throwable t) {
       double cappedFactor = Math.min(Math.max(factor, MIN_FACTOR), MAX_FACTOR);
       double currentRate = limiter.getRate();
       double cappedRate = Math.min(Math.max(currentRate * cappedFactor, MIN_QPS), MAX_QPS);
-      limiter.trySetRate(cappedRate, period);
+      limiter.trySetRate(cappedRate, period, bigtableTracer, cappedFactor, t);
```
mutianf, line 276: I don't think you need the throwable here. The status doesn't mean anything for this metric.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableTracer.java
```diff
@@ -115,4 +114,23 @@ public void grpcMessageSent() {
   public void setTotalTimeoutDuration(Duration totalTimeoutDuration) {
     // noop
   }
+
+  /**
+   * Record the target QPS for batch write flow control.
+   *
+   * @param targetQps The new target QPS for the client.
+   */
+  public void setBatchWriteFlowControlTargetQps(double targetQps) {
```
mutianf, line 123: `@InternalApi`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsConstants.java
```diff
@@ -286,7 +294,23 @@ public static Map<InstrumentSelector, View> getAllViews() {
...
             .add(STREAMING_KEY, STATUS_KEY)
             .build());
-
+    defineView(
+        views,
+        BATCH_WRITE_FLOW_CONTROL_TARGET_QPS_NAME,
+        Aggregation.sum(),
+        InstrumentType.GAUGE,
+        "1",
+        ImmutableSet.<AttributeKey>builder().addAll(COMMON_ATTRIBUTES).add(STATUS_KEY).build());
```
mutianf, line 303: I don't think this is what you defined on the server side? You shouldn't need the status_key

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/RateLimitingServerStreamingCallable.java
```diff
@@ -155,22 +158,26 @@ public double getRate() {
...
+     * @param bigtableTracer The tracer for exporting client-side metrics.
      */
-    public void trySetRate(double rate, Duration period) {
+    public void trySetRate(double rate, Duration period, BigtableTracer bigtableTracer, double cappedFactor,
+                           @Nullable Throwable status) {
       Instant nextTime = nextRateUpdateTime.get();
       Instant now = Instant.now();
 
       if (now.isBefore(nextTime)) {
+        bigtableTracer.addBatchWriteFlowControlFactor(cappedFactor, status, false);
```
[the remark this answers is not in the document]
kongweihan: Hmm good point, I guess in case `context.getTracer() instanceof BigtableTracer` is false? How do we deal with it now? Looks like we don't log error?

Recall that `status` is to record to indicate whether the factor is from the server or due to errors. One theory for the original issue was that a bunch of errors results in `min_factor` and throttles the client. We'd want to confirm that theory and know which error it is in the future.

We want to report this factor distribution even if it's not applied to target QPS, thus the field "applied" in the metric.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsConstants.java
```diff
@@ -286,7 +294,23 @@ public static Map<InstrumentSelector, View> getAllViews() {
...
+    defineView(
+        views,
+        BATCH_WRITE_FLOW_CONTROL_FACTOR_NAME,
+        AGGREGATION_BATCH_WRITE_FLOW_CONTROL_FACTOR_HISTOGRAM,
+        InstrumentType.HISTOGRAM,
+        "1",
+        ImmutableSet.<AttributeKey>builder()
+            .addAll(COMMON_ATTRIBUTES)
+            .add(STREAMING_KEY, STATUS_KEY)
```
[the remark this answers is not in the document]
kongweihan: Ok method is part of common_attributes, and shouldn't I need status_key and applied_key here?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/RateLimitingServerStreamingCallable.java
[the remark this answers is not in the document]
mutianf: We skip exporting the tracer if it's not an instance of BigtableTracer ( which should never happen). I think you can add a sanity check for not null.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableTracer.java
```diff
@@ -115,4 +115,23 @@ public void grpcMessageSent() {
   public void setTotalTimeoutDuration(Duration totalTimeoutDuration) {
     // noop
   }
+
+  /**
+   * Record the target QPS for batch write flow control.
+   *
+   * @param targetQps The new target QPS for the client.
+   */
+  public void setBatchWriteFlowControlTargetQps(double targetQps) {}
```
mutianf, line 124: Please add `@InternalApi` annotation here and the other method.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsConstants.java
```diff
@@ -49,6 +49,7 @@ public class BuiltinMetricsConstants {
   static final AttributeKey<String> METHOD_KEY = AttributeKey.stringKey("method");
   static final AttributeKey<String> STATUS_KEY = AttributeKey.stringKey("status");
   static final AttributeKey<String> CLIENT_UID_KEY = AttributeKey.stringKey("client_uid");
+  static final AttributeKey<Boolean> APPLIED_KEY = AttributeKey.booleanKey("applied");
```
mutianf, line 52: nit: maybe rename it to BATCH_WRITE_QPS_APPLIED_KEY

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsConstants.java
mutianf: It's still here?

mutianf: And it should have method_key

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/RateLimitingServerStreamingCallable.java
```diff
@@ -260,11 +274,11 @@ protected void onCompleteImpl() {
       outerObserver.onComplete();
     }
 
-    private void updateQps(double factor, Duration period) {
+    private void updateQps(double factor, Duration period, @Nullable Throwable t) {
```
mutianf, line 277: Maybe instead of throwable pass in the status. So you don't need to do the null check, you can just have OK status if the update is from a response

kongweihan: But `updateQps` is used by both `onResponseImpl` and `onErrorImpl`

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/RateLimitingServerStreamingCallable.java
```diff
@@ -158,31 +165,49 @@ public double getRate() {
      * @param rate The new rate of the rate limiter.
      * @param period The period during which rate should not be updated again and the rate limiter
      *     should not be disabled.
+     * @param bigtableTracer The tracer for exporting client-side metrics.
      */
-    public void trySetRate(double rate, Duration period) {
+    public void trySetRate(
+        double rate,
+        Duration period,
+        BigtableTracer bigtableTracer,
```
mutianf: `@Nullable` BigtableTracer ..

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/RateLimitingServerStreamingCallable.java
```diff
@@ -260,11 +298,11 @@ protected void onCompleteImpl() {
       outerObserver.onComplete();
     }
 
-    private void updateQps(double factor, Duration period) {
+    private void updateQps(double factor, Duration period, String statusString) {
```
mutianf: why not pass down the throwable?

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTracerTest.java
```diff
@@ -15,22 +15,7 @@
...
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.METHOD_KEY;
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.OPERATION_LATENCIES_NAME;
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.REMAINING_DEADLINE_NAME;
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.RETRY_COUNT_NAME;
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.SERVER_LATENCIES_NAME;
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.STATUS_KEY;
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.STREAMING_KEY;
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.TABLE_ID_KEY;
-import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.ZONE_ID_KEY;
+import static com.google.cloud.bigtable.data.v2.stub.metrics.BuiltinMetricsConstants.*;
```
mutianf: We don't do import .*, this is a setting in intellij: go to Settings/Preferences > Editor > Code Style > Java (or Kotlin) > Imports, and set the "Class count to use import with '*'" field to a very high number (like 999 or 1000)

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsConstants.java
```diff
@@ -20,13 +20,11 @@
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import io.opentelemetry.api.common.AttributeKey;
-import io.opentelemetry.sdk.metrics.Aggregation;
-import io.opentelemetry.sdk.metrics.InstrumentSelector;
-import io.opentelemetry.sdk.metrics.InstrumentType;
-import io.opentelemetry.sdk.metrics.View;
+import io.opentelemetry.sdk.metrics.*;
```
mutianf: here also, no import *

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/Util.java
```diff
@@ -79,7 +79,7 @@ public class Util {
       Metadata.Key.of("x-goog-ext-425905942-bin", Metadata.BINARY_BYTE_MARSHALLER);
 
   /** Convert an exception into a value that can be used to create an OpenCensus tag value. */
-  static String extractStatus(@Nullable Throwable error) {
+  public static String extractStatus(@Nullable Throwable error) {
```
mutianf: this doesn't need to be public now

google-cloud-bigtable/src/test/java/com/google/cloud/bigtable/data/v2/stub/metrics/BuiltinMetricsTestUtils.java
```diff
@@ -91,11 +92,46 @@ public static long getAggregatedValue(MetricData metricData, Attributes attribut
...
+  public static double getAggregatedDoubleValue(MetricData metricData, Attributes attributes) {
+    switch (metricData.getType()) {
+      case HISTOGRAM:
+        HistogramPointData hd =
+            metricData.getHistogramData().getPoints().stream()
+                .filter(pd -> pd.getAttributes().equals(attributes))
+                .collect(Collectors.toList())
+                .get(0);
+        return hd.getSum() / hd.getCount();
+      case LONG_SUM:
```
mutianf: is this right? this is long, how could the value be double?

PR #2695, opened 2025-10-15
feat: Enable ALTS hard bound token in Bigtable w/ direct access

rockspore:


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
```diff
@@ -438,6 +439,10 @@ public static InstantiatingGrpcChannelProvider.Builder defaultGrpcTransportProvi
       grpcTransportProviderBuilder
           .setAttemptDirectPathXds()
           .setAttemptDirectPath(true)
+          // Try to fetch a hard-bound access token for direct access if the runtime
+          // environment supports it.
+          .setAllowHardBoundTokenTypes(
```
sushanb: You can set it as default. Not a hard requirement, but could we add a env var to disable this incase we have any problem? Better to use env var rather than rollback client to prev version.

sushanb (dismissed):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/EnhancedBigtableStubSettings.java
rockspore: This is a great point. Done. Thanks for the tip!

rockspore:
@sushanb Thanks for approving the PR. I doubt the CI failure was due to the change here. Could you help rerun it?

rockspore:
@sushanb Thanks for deflaking the test. Could you help merge it?

PR #2719, opened 2025-11-25
fix(bigtable): add handling for gauge metrics

sushanb:
Field points[0].interval.start_time had an invalid value of "2025-11-25T11:10:06.164418-08:00": The start time must be equal to the end time (2025-11-25T11:11:06.164633-08:00) for the gauge metric 'bigtable.googleapis.com/internal/client/grpc/subchannel/open_connections'

### DIFF
--- a/.kokoro/presubmit/graalvm-native-a.cfg
+++ b/.kokoro/presubmit/graalvm-native-a.cfg
@@ -3,7 +3,7 @@
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
   key: "TRAMPOLINE_IMAGE"
-  value: "gcr.io/cloud-devrel-public-resources/graalvm_sdk_platform_a:3.53.0" # {x-version-update:google-cloud-shared-dependencies:current}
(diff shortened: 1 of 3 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/stub/metrics/BigtableExporterUtils.java
```diff
@@ -281,11 +282,19 @@ private static TimeSeries convertPointToBigtableTimeSeries(
     metricBuilder.putLabels(CLIENT_UID_KEY.getKey(), taskId);
     builder.setMetric(metricBuilder.build());
 
+    MetricKind kind = convertMetricKind(metricData);
+
+    Timestamp endTimestamp = Timestamps.fromNanos(pointData.getEpochNanos());
+    Timestamp startTimestamp;
+
+    if (kind == GAUGE) {
```
mutianf: nit: can we extract this to a helper method? `handleGaugeMetric()` or somehing

PR #2887, opened 2026-04-15
refactor: decouple ConsistencyRequest from TableAdminRequestContext

jinseopkim0:
This commit updates the internal plumbing for `ConsistencyRequest` and `AwaitConsistencyCallable` to support fully qualified table names without breaking the existing public API for the legacy `BigtableTableAdminClient`.

b/502616786


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ConsistencyRequest.java
```diff
@@ -59,17 +61,41 @@ public static ConsistencyRequest forReplication(String tableId, String consisten
...
+  @InternalApi
+  public static ConsistencyRequest forReplicationFromTableName(String tableName) {
+    return new AutoValue_ConsistencyRequest(
+        tableName, CheckConsistencyRequest.ModeCase.STANDARD_READ_REMOTE_WRITES, null, true);
+  }
+
+  @InternalApi
+  public static ConsistencyRequest forReplicationFromTableName(
+      String tableName, String consistencyToken) {
+    Preconditions.checkNotNull(consistencyToken, "consistencyToken must not be null");
```
lqiu96, line 88: Is this something that will be used later? I don't think I see references for this and I'm not entirely sure why there are two variants (one that doesn't take consistencyToken and the other takes the param but also enforces it to be non-null).

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ConsistencyRequest.java
```diff
@@ -83,13 +109,41 @@ public CheckConsistencyRequest toCheckConsistencyProto(
...
+    Preconditions.checkState(
+        isFullyQualified(),
+        "Use toCheckConsistencyProto(TableAdminRequestContext, String) for non-qualified table"
+            + " names.");
+    CheckConsistencyRequest.Builder builder = CheckConsistencyRequest.newBuilder();
+
+    if (getMode().equals(CheckConsistencyRequest.ModeCase.STANDARD_READ_REMOTE_WRITES)) {
+      builder.setStandardReadRemoteWrites(StandardReadRemoteWrites.newBuilder().build());
+    } else {
+      builder.setDataBoostReadLocalWrites(DataBoostReadLocalWrites.newBuilder().build());
```
lqiu96: nit: The logic seems to be mostly the same with the other variant, except for the the logic to build the name. Is it possible for the duplicated logic be extracted to a helper?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ConsistencyRequest.java
```diff
@@ -59,17 +61,41 @@ public static ConsistencyRequest forReplication(String tableId, String consisten
...
+        consistencyToken,
+        true);
   }
 
   @InternalApi
   public CheckConsistencyRequest toCheckConsistencyProto(
       TableAdminRequestContext requestContext, String token) {
+    Preconditions.checkState(
+        !isFullyQualified(),
+        "Use toCheckConsistencyProto(String token) for fully qualified table names.");
```
lqiu96, line 119: qq, `toCheckConsistencyProto(String token)` is the non-fully qualified table right? IIUC, this should be the inverse?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/EnhancedBigtableTableAdminStub.java
```diff
@@ -54,14 +54,19 @@ public class EnhancedBigtableTableAdminStub extends GrpcBigtableTableAdminStub {
...
   @Deprecated private final AwaitReplicationCallable awaitReplicationCallable;
 
   private final AwaitConsistencyCallable awaitConsistencyCallable;
   private final OperationCallable<Void, Empty, OptimizeRestoredTableMetadata>
       optimizeRestoredTableOperationBaseCallable;
 
+  public static EnhancedBigtableTableAdminStub createEnhanced(
+      BigtableTableAdminStubSettings settings) throws IOException {
+    return new EnhancedBigtableTableAdminStub(settings, ClientContext.create(settings), null);
+  }
```
lqiu96, line 77: Is it possible to add javadoc for this new public method?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ConsistencyRequest.java
jinseopkim0: Yes, this will be used in the follow-up PR https://github.com/googleapis/java-bigtable/pull/2889. 

Note: This followed the structure for `forReplication()`.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ConsistencyRequest.java
jinseopkim0: Thanks for the feedback. I've created a helper.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ConsistencyRequest.java
jinseopkim0: `toCheckConsistencyProto(String token)` is the fully qualified table. I've added comments to make that clearer.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/EnhancedBigtableTableAdminStub.java
jinseopkim0: Thanks for the feedback, done.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ConsistencyRequest.java
```diff
@@ -59,37 +61,106 @@ public static ConsistencyRequest forReplication(String tableId, String consisten
...
+        tableName, CheckConsistencyRequest.ModeCase.STANDARD_READ_REMOTE_WRITES, null, true);
+  }
+
+  @InternalApi
+  public static ConsistencyRequest forReplicationFromTableName(
+      String tableName, String consistencyToken) {
+    Preconditions.checkNotNull(consistencyToken, "consistencyToken must not be null");
+
+    return new AutoValue_ConsistencyRequest(
+        tableName,
```
lqiu96, line 91: (just a general question): If this uses the fully qualified name (i assume by the true param), do we need to do any validation to ensure that the input is fully qualified?

What happens if isn't?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/EnhancedBigtableTableAdminStub.java
```diff
@@ -54,14 +54,28 @@ public class EnhancedBigtableTableAdminStub extends GrpcBigtableTableAdminStub {
...
+   *
+   * @param settings The settings used to configure the stub.
+   * @return A new instance of {@code EnhancedBigtableTableAdminStub}.
+   * @throws IOException If there are errors creating the underlying client context.
+   */
+  public static EnhancedBigtableTableAdminStub createEnhanced(
+      BigtableTableAdminStubSettings settings) throws IOException {
+    return new EnhancedBigtableTableAdminStub(settings, ClientContext.create(settings), null);
+  }
+
```
lqiu96, line 78: nit: Not something that needs to be done in this PR. Thoughts on adding a similar one for the `createEnhanced` below? I think it would be helpful to help users know when to pick this one vs the other? 

e.g. are there certain cases where they don't know the fully qualified `TableAdminRequestContext` values? Or perhaps don't want it?

lqiu96 (approved):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/models/ConsistencyRequest.java
jinseopkim0: Thanks for the feedback. I've added validations to ensure that the input is fully qualified.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/EnhancedBigtableTableAdminStub.java
jinseopkim0: Thanks for the feedback. I've added javadoc for the method below.

PR #2889, opened 2026-04-15
feat: introduce Bigtable Admin V2 clients for Selective GAPIC

jinseopkim0:
This commit adds `BigtableTableAdminClientV2` and `BigtableInstanceAdminClientV2`, which extend the autogenerated base clients. These classes serve as the new, independent entry points for the Selective GAPIC hierarchy.

For `BigtableTableAdminClientV2`, the manual wrappers for Critical User Journeys (CUJs) like `awaitOptimizeRestoredTable` and `waitForConsistency` have been relocated from the legacy client. It utilizes the new context-free ConsistencyRequest factories introduced in the previous refactoring step.

For `BigtableInstanceAdminClientV2`, the class acts as a standardized entry point, as no manual overrides were required from the legacy client.

b/502616786


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
```diff
@@ -0,0 +1,182 @@
...
+  protected BigtableTableAdminClientV2(BaseBigtableTableAdminSettings settings) throws IOException {
+    super(settings);
+  }
+
+  protected BigtableTableAdminClientV2(BigtableTableAdminStub stub) {
+    super(stub);
+  }
+
+  /** Constructs an instance of BigtableTableAdminClientV2 with the given settings. */
+  public static final BigtableTableAdminClientV2 createClient(
```
igorbernstein2, line 197: Lets keep this consistent with the create factory methods on BaseBigtableTableAdminClient

ie s/createClient/create

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
```diff
@@ -0,0 +1,182 @@
...
+      BaseBigtableTableAdminSettings settings) throws IOException {
+    // Explicitly create the enhanced stub
+    EnhancedBigtableTableAdminStub stub =
```
igorbernstein2: why is this referencing the EnhancedStub? The V2 stack should be independent of the old manual wrappers

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
jinseopkim0: Using `create` causes errors like these:
```
[ERROR] ... BigtableTableAdminClientV2.java:[197,49] error: create(BaseBigtableTableAdminSettings) in BigtableTableAdminClientV2 cannot override create(BaseBigtableTableAdminSettings) in
     BaseBigtableTableAdminClient
[ERROR]   overridden method is static,final
```

`BaseBigTableTableAdminClient` has the `public static final` prefixes: https://github.com/googleapis/java-bigtable/blob/0aecd6aa7e27a71dcfddb3255c4663000839f05c/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BaseBigtableTableAdminClient.java#L881

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
jinseopkim0: Thanks for catching that. I've updated the code.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
igorbernstein2: Can we chat with gapic generator folks about removing the final modifier from create? because we explicitly want to hide the underlying create methods

jinseopkim0: Thanks for the feedback. Started a chat thread separately (CC @blakeli0, @lqiu96).

jinseopkim0: Created https://github.com/googleapis/java-bigtable/pull/2898 to use the postprocessor to remove the `final` modifier.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientV2.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.google.cloud.bigtable.admin.v2.stub.BigtableInstanceAdminStub;
+import java.io.IOException;
+
+/**
+ * Modern Cloud Bigtable Instance Admin Client.
+ *
+ * <p>This client extends the auto-generated {@link BaseBigtableInstanceAdminClient} to provide
+ * manual overrides and additional convenience methods for Critical User Journeys (CUJs) that the
+ * GAPIC generator cannot handle natively.
```
lqiu96: nit: I don't think this part needs to be in the javadoc as these terms probably have no meaning to users. Perhaps this section could be modified to explain a bit more about
- chained LROs and how that benefits the UX/ user
- Consistency polling and why it's been improved over the old legacy version

same with the tableadminclient

WDYT?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
```diff
@@ -0,0 +1,338 @@
...
+    // TODO(igorbernstein2): expose polling settings
+    RetrySettings pollingSettings =
+        RetrySettings.newBuilder()
+            .setTotalTimeout(
+                settings.checkConsistencySettings().getRetrySettings().getTotalTimeout())
+            .setInitialRetryDelay(Duration.ofSeconds(10))
+            .setRetryDelayMultiplier(1.0)
+            .setMaxRetryDelay(Duration.ofSeconds(10))
+            .setInitialRpcTimeout(Duration.ZERO)
+            .setMaxRpcTimeout(Duration.ZERO)
```
lqiu96: can we use the java.time variants? It would be easier to migrate and deprecate off of threetenbp variants in the future

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientV2.java
jinseopkim0: Thanks for the feedback. I've updated the javadoc as suggested. Please let me know if you have any additional feedback.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
jinseopkim0: Thanks for the feedback, done.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientV2.java
```diff
@@ -0,0 +1,48 @@
...
+  /** Constructs an instance of BigtableInstanceAdminClientV2 with the given settings. */
+  public static final BigtableInstanceAdminClientV2 create(
+      BaseBigtableInstanceAdminSettings settings) throws IOException {
+    return new BigtableInstanceAdminClientV2(settings);
+  }
+
+  /** Constructs an instance of BigtableInstanceAdminClientV2 with the given stub. */
+  public static final BigtableInstanceAdminClientV2 create(BigtableInstanceAdminStub stub) {
+    return new BigtableInstanceAdminClientV2(stub);
+  }
```
lqiu96, line 47: qq, what is the difference between these two? I asssume one is for the old stub?

lqiu96 (approved):
Changes overall LGTM.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableInstanceAdminClientV2.java
jinseopkim0: Actually, neither is for the old stub, as the new clients do not depend on the legacy clients. The methods are different possible initialization paths. The first (with the `BaseBigtableInstanceAdminSettings`) will be the standard path, and the second (with the `BigtableInstanceAdminStub`) will allow passing an already constructed stub. Note that they mirror the factory methods in https://github.com/googleapis/java-bigtable/blob/77952d2f4ae32da23ed258955038e0756b2ea3f0/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BaseBigtableInstanceAdminClient.java#L789.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
```diff
@@ -0,0 +1,341 @@
...
+      OperationCallable<Void, Empty, OptimizeRestoredTableMetadata>
+          optimizeRestoredTableOperationBaseCallable) {
+    super(stub);
+    this.awaitConsistencyCallable = awaitConsistencyCallable;
+    this.optimizeRestoredTableOperationBaseCallable = optimizeRestoredTableOperationBaseCallable;
+  }
+
+  private AwaitConsistencyCallable createAwaitConsistencyCallable(
+      BigtableTableAdminStubSettings settings) throws IOException {
+    ClientContext clientContext = ClientContext.create(settings);
```
igorbernstein2: This is not great, ClientContexts are pretty heavy (they create a new grpc channel). I think the only reason you need from the context is the clock & executor. I think you'll be better off creating those for this child class and not trying getting access to the ClientContext

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
```diff
@@ -0,0 +1,341 @@
...
+    // TODO(igorbernstein2): expose polling settings
+    RetrySettings pollingSettings =
```
igorbernstein2, line 135: Lets extract this to a constant

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
```diff
@@ -0,0 +1,341 @@
...
+      createOptimizeRestoredTableOperationBaseCallable(BigtableTableAdminStubSettings settings)
+          throws IOException {
+    ClientContext clientContext = ClientContext.create(settings);
+
+    GrpcCallSettings<Void, Operation> unusedInitialCallSettings =
+        GrpcCallSettings.create(
+            MethodDescriptor.<Void, Operation>newBuilder()
+                .setType(MethodType.UNARY)
+                .setFullMethodName(
+                    "google.bigtable.admin.v2.BigtableTableAdmin/OptimizeRestoredTable")
```
igorbernstein2: I wonder if it would be better to declare a dep on the grpc stubs and pull this from the MethodDescriptor?

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
jinseopkim0: Thanks for the comment. I've updated the code as suggested. Please let me know if you have any other feedback / concerns.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
jinseopkim0: Thanks for the feedback, done.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
jinseopkim0: Thanks for the feedback. I think it improves readability, so I've updated code.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/AwaitConsistencyCallable.java
```diff
@@ -51,15 +51,15 @@
  * <p>This callable wraps GenerateConsistencyToken and CheckConsistency RPCs. It will generate a
  * token then poll until isConsistent is true.
  */
-class AwaitConsistencyCallable extends UnaryCallable<ConsistencyRequest, Void> {
+public class AwaitConsistencyCallable extends UnaryCallable<ConsistencyRequest, Void> {
```
igorbernstein2: I think this needs to be marked InternalApi ... I dont think we actually expose this to end users

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/AwaitConsistencyCallable.java
```diff
@@ -51,15 +51,15 @@
...
   private final UnaryCallable<GenerateConsistencyTokenRequest, GenerateConsistencyTokenResponse>
       generateCallable;
   private final UnaryCallable<CheckConsistencyRequest, CheckConsistencyResponse> checkCallable;
   private final RetryingExecutor<CheckConsistencyResponse> executor;
 
   @Nullable private final TableAdminRequestContext requestContext;
 
-  static AwaitConsistencyCallable create(
+  public static AwaitConsistencyCallable create(
```
igorbernstein2: Same as bove

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/AwaitConsistencyCallable.java
```diff
@@ -79,7 +79,7 @@ static AwaitConsistencyCallable create(
         generateCallable, checkCallable, retryingExecutor, requestContext);
   }
 
-  static AwaitConsistencyCallable create(
+  public static AwaitConsistencyCallable create(
```
igorbernstein2: same as above

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
```diff
@@ -0,0 +1,377 @@
...
+    super(settings);
+    // Extract the executor directly without spinning up a full channel
+    this.backgroundExecutor =
+        settings.getStubSettings().getBackgroundExecutorProvider().getExecutor();
+    com.google.api.gax.rpc.ClientContext lightweightContext =
+        com.google.api.gax.rpc.ClientContext.newBuilder()
+            .setClock(settings.getStubSettings().getClock())
+            .setExecutor(this.backgroundExecutor)
+            .setDefaultCallContext(com.google.api.gax.grpc.GrpcCallContext.createDefault())
+            .build();
```
igorbernstein2: we should talk

igorbernstein2 (changes requested):

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/AwaitConsistencyCallable.java
jinseopkim0: Thanks for the feedback, done.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/AwaitConsistencyCallable.java
jinseopkim0: Thanks for the feedback, done.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/stub/AwaitConsistencyCallable.java
jinseopkim0: Thanks for the feedback, done.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
jinseopkim0: Discussed offline, aligned on approach to remove `ClientContext` usage completely. I've updated code accordingly.

PR #2896, opened 2026-04-20
fix: Update renovate config check to use npx

suztomo:
This PR updates the renovate config check workflow to use npx instead of global installation, avoiding issues with missing modules.


igorbernstein2 (approved):

suztomo:
cloud-java-bot reverted the change.

PR #2897, opened 2026-04-21
fix: cancel even when vrpc is not started yet and fix newRealCall

mutianf:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/session/VRpcImpl.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/session/VRpcImpl.java
@@ -20,6 +20,7 @@ import com.google.bigtable.v2.PeerInfo;
 import com.google.bigtable.v2.VirtualRpcRequest;
 import com.google.bigtable.v2.VirtualRpcRequest.Metadata;
 import com.google.bigtable.v2.VirtualRpcResponse;
+import com.google.cloud.bigtable.data.v2.internal.csm.tracers.DebugTagTracer;
 import com.google.cloud.bigtable.data.v2.internal.middleware.VRpc;
 import com.google.protobuf.Message;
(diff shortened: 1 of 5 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/session/VRpcImpl.java
```diff
@@ -124,9 +131,11 @@ public void start(ReqT req, VRpcCallContext ctx, VRpcListener<RespT> listener) {
     }
 
     if (!status.isOk()) {
-      if (!state.compareAndSet(State.STARTED, State.CLOSED)) {
-        return;
```
nimf, line 128: The return is missing now. It was here to protect from the race with handleSessionClose, right? Then we should keep it.

mutianf: it'll now throw an exception if the state is not started. see here: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/tracers/DebugTagTracer.java#L27

nimf: oh. Do we want a user to have an exception if starting a call coincides with session closing? Am I missing something?

mutianf: Hmm, I think this only prevents double start of an vrpc which should never happen? session closing should be handled here: https://github.com/googleapis/java-bigtable/blob/main/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/session/VRpcImpl.java#L140 which we only log a warning if the state is not correct.

igorbernstein2 (approved):

PR #2898, opened 2026-04-21
build(owlbot): un-final static create methods in Admin Base clients

jinseopkim0:
Updates owlbot.py to programmatically remove the \`final\` modifier from the static \`create\` factory methods in the generated BaseBigtableTableAdminClient and BaseBigtableInstanceAdminClient.

b/502616786


owlbot.py
```diff
@@ -67,6 +67,14 @@ def make_internal_only(sources):
   if os.path.exists("owl-bot-staging/v2/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/package-info.java"):
     os.remove("owl-bot-staging/v2/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/package-info.java")
 
+  # Remove the 'final' modifier from all static create() factory methods in the Base Admin clients
+  # This allows our handwritten Selective GAPIC subclasses to hide these methods.
+  s.replace(
+      f"{library}/**/BaseBigtable*AdminClient.java",
+      r"public static final BaseBigtable(.*)AdminClient create\(",
+      r"public static BaseBigtable\1AdminClient create("
```
blakeli0: Can we remove `public` as well if they are supposed to be overridden?

jinseopkim0: Thanks for the feedback, done.

blakeli0 (approved):

PR #2902, opened 2026-04-23
fix(bigtable): Use a different t4t7 latencies for directpath

sushanb:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/tracers/BuiltinMetricsTracer.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/tracers/BuiltinMetricsTracer.java
@@ -400,7 +400,8 @@ class BuiltinMetricsTracer extends BigtableTracer {
           code,
           Comparators.max(remainingDeadlineAtAttemptStart, Duration.ZERO));
     }
-
+    // if we don't have t4t7 latency from gfe, we use dur between initial metadata sent and initial
+    // metadata recv
     if (sidebandData.getGfeTiming() != null) {
       recorder.serverLatency.record(
           clientInfo,
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/tracers/BuiltinMetricsTracer.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/tracers/BuiltinMetricsTracer.java
@@ -271,7 +271,7 @@ class BuiltinMetricsTracer extends BigtableTracer {
 
   @Override
   public void grpcHeadersReceived() {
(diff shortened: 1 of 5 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/tracers/BuiltinMetricsTracer.java
```diff
@@ -82,6 +82,8 @@ class BuiltinMetricsTracer extends BigtableTracer {
   private final AtomicLong totalClientBlockingTime = new AtomicLong(0);
 
   private final AtomicLong grpcMessageSentDelay = new AtomicLong(0);
+  private final AtomicLong grpcHeadersSentNanos = new AtomicLong(0);
```
igorbernstein2: Why not a volatile Stopwatch? it would communicate the intent better:


```java
  volatile Optional<Stopwatch> faket4t7 = Optional.empty();

public void grpcHeadersSent() {
  faket4t7 = Optional.of(Stopwatch.createStarted());
}

public void grpcHeadersReceived() {
  faket4t7.map(Stopwatch::stop);
}

recordAttemptCompletion() {
  if (sidebandData.getGfeTiming() != null) {
  } else if (faket4t7.isPresent()) {
     ....
  }
}
....
```

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/tracers/BuiltinMetricsTracer.java
```diff
@@ -395,6 +410,18 @@ private void recordAttemptCompletion(@Nullable Throwable throwable) {
           sidebandData.getClusterInfo(),
           code,
           sidebandData.getGfeTiming());
+    } else {
```
igorbernstein2, line 410: This is expanding scope to what we talked about. This was supposed to be only for DirectAccess. This implementation expands the scope to any time the header is missing. Which means that we wont be able to tell what we are actually measuring. In our original conversation scoping it only to DA, we could reason that when transport_type is DA, we measuring client headers until response headers. In all other cases we are measuring gfe latency. And if the DistributionCount() drops we know that we couldnt reach a gfe. With this approach, we muddy our ability to reason about the metric

sushanb: Don't we have connectivity error count for measuring that particular usecase you are talking about?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/csm/tracers/BuiltinMetricsTracer.java
sushanb: Done.

PR #2903, opened 2026-04-24
feat: recycle channel on consecutive new stream failures

nimf:

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/channels/ChannelPoolDpImpl.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/channels/ChannelPoolDpImpl.java
@@ -65,6 +65,8 @@ public class ChannelPoolDpImpl implements ChannelPool {
   private static final String DEFAULT_LOG_NAME = "pool";
   private static final AtomicInteger INDEX = new AtomicInteger();
 
+  private static final int CONSECUTIVE_OPEN_FAILURE_THRESHOLD = 5;
+
   private final String poolLogId;
 
   @VisibleForTesting volatile int minGroups;
@@ -221,6 +223,7 @@ public class ChannelPoolDpImpl implements ChannelPool {
               public void onBeforeSessionStart(PeerInfo peerInfo) {
                 afeId = AfeId.extract(peerInfo);
                 synchronized (ChannelPoolDpImpl.this) {
+                  channelWrapper.consecutiveFailures = 0;
                   rehomeChannel(channelWrapper, afeId);
                   sessionsPerAfeId.add(afeId);
                 }
@@ -232,6 +235,8 @@ public class ChannelPoolDpImpl implements ChannelPool {
                 synchronized (ChannelPoolDpImpl.this) {
                   if (afeId != null) {
                     sessionsPerAfeId.remove(afeId);
+                  } else if (!status.isOk() && status.getCode() != Code.CANCELLED) {
+                    channelWrapper.consecutiveFailures++;
                   }
                   releaseChannel(channelWrapper, status);
                 }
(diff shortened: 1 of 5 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/channels/ChannelPoolDpImpl.java
```diff
@@ -232,6 +235,8 @@ public void onClose(Status status, Metadata trailers) {
                 synchronized (ChannelPoolDpImpl.this) {
                   if (afeId != null) {
                     sessionsPerAfeId.remove(afeId);
+                  } else if (!status.isOk() && status.getCode() != Code.CANCELLED) {
+                    channelWrapper.consecutiveFailures++;
                   }
                   releaseChannel(channelWrapper, status);
```
mutianf, line 249: Not related to this PR, but I feel releaseChannel should be called releaseStream or something, with recycleChannel the naming gets a little confusing..

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/channels/ChannelPoolDpImpl.java
```diff
@@ -306,12 +311,12 @@ private void releaseChannel(ChannelWrapper channelWrapper, Status status) {
...
     channelWrapper.numOutstanding--;
 
-    if (shouldRecycleChannel(status)) {
+    if (shouldRecycleChannel(channelWrapper, status)) {
       recycleChannel(channelWrapper);
     }
   }
 
-  private static boolean shouldRecycleChannel(Status status) {
+  private static boolean shouldRecycleChannel(ChannelWrapper channelWrapper, Status status) {
```
mutianf, line 327: Also unrelated to this PR.. I'm wondering how this works..

if we get unimplemented, shouldn't the client fallback to unary? (iiuc which doesn't use this channel pool) recycling the channel and try to create a new one likely still won't work?

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/channels/ChannelPoolDpImpl.java
```diff
@@ -65,6 +65,8 @@ public class ChannelPoolDpImpl implements ChannelPool {
   private static final String DEFAULT_LOG_NAME = "pool";
   private static final AtomicInteger INDEX = new AtomicInteger();
 
+  private static final int CONSECUTIVE_OPEN_FAILURE_THRESHOLD = 5;
```
mutianf: wondering if this should be called CONSECUTIVE_OPEN_SESSION_FAILURE_THRESHOLD.

And I think this should come from client config?

nimf: Renamed and added a todo.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/channels/ChannelPoolDpImpl.java
nimf: Yes, we will eventually fallback to unary if recycling doesn't help. But I've added a backoff not to spam new connections rapidly and indefinitely.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/data/v2/internal/channels/ChannelPoolDpImpl.java
nimf: I guess the meaning here was that a session is closing and releasing the channel it was using.

PR #2904, opened 2026-04-24
refactor(samples): migrate Admin API samples to BigtableAdminClientV2

jinseopkim0:
This commit modernizes the Bigtable Admin API samples to use the V2 Selective GAPIC clients (`BigtableTableAdminClientV2` and `BigtableInstanceAdminClientV2`).

### DIFF
--- a/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
+++ b/google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
@@ -56,7 +56,8 @@ import org.threeten.bp.Duration;
  * generator cannot handle natively (e.g., chained Long Running Operations, Consistency Polling).
  */
 public class BigtableTableAdminClientV2 extends BaseBigtableTableAdminClient {
-  private final com.google.cloud.bigtable.admin.v2.stub.AwaitConsistencyCallable awaitConsistencyCallable;
+  private final com.google.cloud.bigtable.admin.v2.stub.AwaitConsistencyCallable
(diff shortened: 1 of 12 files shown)


google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
lqiu96: nit: Perhaps you can link to the setter? It's not immediately obvious what I need to change to enable this (from the perspective of someone that's not really touched bigtable)

lqiu96:
Changes LGTM. QQ, is there a sample that shows users how to migrate from the legacy client to the new V2 client? Or is there really nothing that users need to do?

jinseopkim0:
> Changes LGTM. QQ, is there a sample that shows users how to migrate from the legacy client to the new V2 client? Or is there really nothing that users need to do?

https://docs.cloud.google.com/bigtable/docs/upgrading-clients has information on the migration. This will also be updated once PRs are merged.

google-cloud-bigtable/src/main/java/com/google/cloud/bigtable/admin/v2/BigtableTableAdminClientV2.java
jinseopkim0: Thanks for the feedback. This change was actually not supposed to be part of this PR (rebase out-of-sync is showing these diffs). I've updated https://github.com/googleapis/java-bigtable/pull/2889 to make the error clearer.

lqiu96:
PTAL. I think there is a compile error in the CI, but the changes overall LGTM

